PR #6, opened 2018-10-22
Conner/daily fetch

ilikesymmetry:
Added a feature that allows the server to get and parse Cornell Dining's JSON and update our graphql db every set time period.  The current set time period is 24 hours.

### DIFF
--- a/constants.py
+++ b/constants.py
@@ -1,0 +1,1 @@
+UPDATE_DELAY = 86400  # 24 hours in seconds
--- a/data.py
+++ b/data.py
@@ -1,8 +1,9 @@
-# Will update the graphQL db by calling scraper.py every defined time interval
-from datetime import datetime
-import requests
-import schema
 from bs4 import BeautifulSoup
+import constants
+from datetime import datetime
+from requests import get
+import schema
+from threading import Timer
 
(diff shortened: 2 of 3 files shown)


constants.py
```diff
@@ -0,0 +1 @@
+UPDATE_DELAY = 86400 #24hrs
```
young-k: 24 hours in milliseconds (make the comment more clear), also there should be a space after the `#`, so `# 24`

young-k (approved):
lgtm, make changes before merging

data.py
```diff
@@ -1,8 +1,10 @@
 # Will update the graphQL db by calling scraper.py every defined time interval
 from datetime import datetime
-import requests
+from requests import get
 import schema
 from bs4 import BeautifulSoup
+from threading import Timer
```
young-k, line 6: alphabetize these imports, so `bs4` at the top, ... etc

young-k (changes requested):
jk, one more round

ilikesymmetry:
@young-k

young-k (approved):
In the future, make sure to change the pull request title (should be something like `Added daily fetch...`).

Also rebase away the last two commits before merging!

PR #11, opened 2018-10-28
Add Dockerfile

meganle:
To run it locally, build it
`docker build -t eatery .`
and then to run it
`docker run -d -p 5000:5000 eatery`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,4 +9,4 @@ RUN pip install -r requirements.txt
 
 EXPOSE 5000 
 
-CMD flask run --host=0.0.0.0
+CMD python app.py 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.6
+
+RUN mkdir /usr/src/app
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN mkdir /usr/src/app
+WORKDIR /usr/src/app
+
+COPY . .
+
+RUN pip install -r requirements.txt
+
+EXPOSE 5000 
+
+CMD flask run --host=0.0.0.0
```
kq-li: prefer `python app.py`, think this does some work under the hood that we may not want (env vars) -- though we'll probably do gunicorn soon anyway

kq-li (approved):
lgtm, maybe `flask run` => `python app.py` for transparency

kq-li:
also you may not want to daemonize the container (-d flag) when running it so you can actually see output and such, probably best to just run it on a separate terminal for local testing purposes

young-k:
LGTM, check out kenneth's comments (insightful as always)

PR #19, opened 2018-11-06
Conversion from DiningStack external eateries to GraphQL

ilikesymmetry:
Added a missing key handler for keys that statics eateries may not have.  Created mechanism for taking external eatery hours and convert to our current graphql style.  Added try-excepts to handle missing information in external eateries. #18


data.py
```diff
@@ -1,31 +1,25 @@
-import constants
-from datetime import datetime
+from constants import *
```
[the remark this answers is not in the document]
young-k: @cswenberg hound is correct, try and avoid `*` imports and just name all the imports instead

data.py
```diff
@@ -137,33 +131,128 @@ def parse_food_items(item_list, eatery_id):
...
+    for i in range(100):
+      test_id = 100 - i
+      if test_id not in eateries:
+        return test_id
+
+def parse_eatery_type(eatery):
+  try:
+    return eatery['eateryTypes'][0]['descr']
+  except Exception:
+    return 'unknown'
```
young-k: better exception message, we're gonna add logging to these in the future

data.py
```diff
@@ -137,33 +131,128 @@ def parse_food_items(item_list, eatery_id):
...
+      print('weekday key: ' + hours['weekday'] + ' not in constants WEEKDAYS!')
+      continue
+    for weekday in WEEKDAYS[hours['weekday']]:
+      if weekday not in weekdays:
+        weekdays[weekday] = hours['events']
+
+  global operating_hours
+  global today
+  new_operating_hours = []
+  for i in range(8):
```
young-k: why 8, and can you define this in a constant at the top?

data.py
```diff
@@ -137,33 +131,128 @@ def parse_food_items(item_list, eatery_id):
...
+    new_date = date(today.year, today.month, today.day + i)
+    weekday = new_date.weekday()
+    new_events = weekdays[weekday] if weekday in weekdays.keys() else []
```
young-k: this should be: `new_events = weekdays.get(weekday, [])`

data.py
```diff
@@ -137,33 +131,128 @@ def parse_food_items(item_list, eatery_id):
...
+def safe_parse(obj, key, default=''):
+  try:
+    value = obj[key]
+    if value is None:
+      return default
+    return value
+  except Exception:
+    return default
+
+def resolve_id(obj):
```
young-k: can we add comments to all of these functions explaining what they do, ex:

```
def resolve_id(obj):
  """Returns [SOMETHING]
  [OPTIONAL EXPLANATION IF NECESSARY]

  Args:
    obj: [INSERT HERE]
  """
```

Refer to https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods

young-k (changes requested):
Pretty bulky changes, try and get these in by Wednesday!

data.py
ilikesymmetry: iOS uses an enum from the value for 'eateryType' where unknown is a valid value when we can't determine another value otherwise

ilikesymmetry:
@young-k changes made

data.py
```diff
@@ -1,31 +1,26 @@
-import constants
-from datetime import datetime
+from constants import ASSET_BASE_URL, CORNELL_DINING_URL, NUM_DAYS_STORED, STATIC_EATERIES_URL, \
+    PAY_METHODS, UPDATE_DELAY, WEEKDAYS
```
kq-li: for long imports, prefer
```
from constants import (
    ASSET_BASE_URL,
    CORNELL_DINING_URL,
    ...
)
```

or just `from constants import *`

data.py
```diff
@@ -62,7 +57,7 @@ def parse(data_json):
     eateries[new_eatery.id] = new_eatery
 
 def get_image_url(slug):
-  return constants.ASSET_BASE_URL + 'eatery-images/' + slug + '.jpg'
+  return ASSET_BASE_URL + 'eatery-images/' + slug + '.jpg'
```
kq-li: refactor this, `IMAGES_URL` perhaps?

kq-li: alternatively, consider a utils class with a function to generate an image URL from a given name

data.py
```diff
@@ -137,33 +132,133 @@ def parse_food_items(item_list, eatery_id):
...
+        name=safe_parse(eatery, 'name'),
+        name_short=safe_parse(eatery, 'nameshort'),
+        operating_hours=parse_static_op_hours(eatery['operatingHours'], new_id),
+        payment_methods=parse_payment_methods(eatery['payMethods']),
+        phone=parse_phone(eatery['contactPhone'], eatery['name']),
+        slug=safe_parse(eatery, 'slug')
+    )
+    eateries[new_eatery.id] = new_eatery
+
+def safe_parse(obj, key, default=''):
```
kq-li: A++ intuition, the developers of python agreed with you that this is useful and bestowed us with `dict.get(key, default)` which is pretty much the best thing ever invented

data.py
```diff
@@ -137,33 +132,133 @@ def parse_food_items(item_list, eatery_id):
...
+    return value
+  except Exception:
+    return default
+
+def resolve_id(eatery):
+  """Returns a new id (int) for an external eatery
+  If the eatery does not have a provided id, we need to make one for it.
+  Works by going backwards from 99 and making sure the id is not already takenself.
+  Should feasibly never conflict with cornell's eateries (numbered ascending, current max is < 30).
+  """
```
kq-li, line 211: prefer triple single quotes `'''` over doubles `"""`

data.py
```diff
@@ -137,33 +132,133 @@ def parse_food_items(item_list, eatery_id):
...
+  try:
+    id = eatery['id']
+    return id
+
+  except Exception:
+    global eateries
+    for i in range(100):
+      test_id = 100 - i
+      if test_id not in eateries:
+        return test_id
```
kq-li: is this for debugging? what's the significance of 100?

data.py
```diff
@@ -137,33 +131,128 @@ def parse_food_items(item_list, eatery_id):
...
+def parse_eatery_type(eatery):
+  try:
+    return eatery['eateryTypes'][0]['descr']
+  except Exception:
+    return 'unknown'
+
+def parse_static_op_hours(hours_list, eatery_id):
+  weekdays = {}
+  for hours in hours_list:
+    if hours['weekday'] not in WEEKDAYS.keys():
```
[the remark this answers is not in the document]
kq-li: `x not in dict.keys()` and `x not in dict` are equivalent in python

kq-li (commented):
Good stuff -- just some minor fixes needed, but otherwise lgtm

constants.py
```diff
@@ -1,3 +1,25 @@
...
+WEEKDAYS = {
+    'monday-wednesday': [0, 1, 2],
+    'monday-thursday': [0, 1, 2, 3],
+    'monday-friday': [0, 1, 2, 3, 4],
+    'thursday': [3],
+    'friday': [4],
+    'saturday': [5],
+    'sunday': [6]
+}
+NUM_DAYS_STORED = 8
```
alanna-zhou: why is this 8?

ilikesymmetry: Cornell Dining's json stores today and the next week of eatery information, so our graphql db stores 8 days worth of information at any point in time

data.py
young-k: @kq-li this isn't correct, docstrings are in double quotes

young-k: @cswenberg can you redo this change?

data.py
```diff
@@ -129,41 +128,125 @@ def parse_food_items(item_list, eatery_id):
...
+  try:
+    return eatery['eateryTypes'][0]['descr']
+  except Exception:
+    return 'Unknown'
+
+def parse_static_op_hours(hours_list, eatery_id):
+  weekdays = {}
+  for hours in hours_list:
+    if hours['weekday'] not in WEEKDAYS:
+      print('weekday key: ' + hours['weekday'] + ' not in constants WEEKDAYS!')
```
young-k: use a format string here

data.py
```diff
@@ -129,41 +128,125 @@ def parse_food_items(item_list, eatery_id):
...
+        phone=eatery.get('contactPhone', 'N/A'),
+        slug=eatery.get('slug', '')
+    )
+    eateries[new_eatery.id] = new_eatery
+
+def resolve_id(eatery):
+  '''Returns a new id (int) for an external eatery
+  If the eatery does not have a provided id, we need to make one for it.
+  Simply gets the maximum id currently in our eateries and makes new id +1
+  '''
```
young-k: change this back to double quotes

constants.py
young-k: make this more verbose, `NUM_DAYS_STORED_IN_DB`

young-k: alphabetize this file as well

young-k (changes requested):
Almost there, minor changes this time, will approve afterwards.

data.py
young-k: @kq-li we're trying to avoid `*` imports, it's not good practice in general

young-k:
Another comment, please press the `Resolve` button on comments after you resolve them!

data.py
kq-li: whoops, you're right, my bad

young-k (approved):
Great job with this! *Try* rebasing and squash the previous commits to fix PR stuff, if it doesn't work just `rebase --abort` and merge in!

PR #20, opened 2018-11-07
Add gunicorn

ilikesymmetry:
#8


young-k (changes requested):
Can you change the Dockerfile to reflect this as well?

young-k (approved):
LGTM

PR #24, opened 2018-11-12
 Change return time from UTC to EDT

henrytso:


young-k (approved):
LGTM, it's `EST` and not `EDT` though

young-k:
Also – the branch name here didn't correspond to what you were working on, make sure that matches up in the future!

young-k (changes requested):
Wait, rebase the message of the second commit, you spelled `requirements` wrong –– change the first one from `EDT` to `EST` too!

`git rebase -i HEAD~2` is what you want, and reword both commits!

PR #27, opened 2018-11-12
Prevent null EateryType phone field

young-k:
Added some print statements, found out that sometimes the `value` in the dictionary is `None`, so added a ternary to avoid that case.


kevinchan159 (approved):
awesome work Young!

young-k:
Addressed Hound comment through `rebase`.

PR #35, opened 2018-11-12
Fix bug where static items were not showing up

young-k:
Fixed (Verified that this didn't break any current functionality):
screen shot 2018-11-12 at 10 45 20 am


AAAstorga (approved):
LGTM

PR #37, opened 2018-11-13
Major refactoring of types

young-k:
The project is currently architected as follows:
```
src 
|-- schema.py
|-- data.py
```

`schema.py` was bloated with both resolvers and all types, so I changed the architecture to be the following:

```
src 
|-- schema.py
|-- data.py
|-- types/
     |-- __init__.py
     |-- account_info_type.py
     |-- eatery_type.py
     |-- food_station_type.py
     |-- operating_hours_type.py
     |-- payment_methods_type.py
```

I wanted to avoid having all the classes in one file, while avoiding having a file per each class (since each is only a few lines), so I tried to group by functionality.

I also verified that no functionality broke after this refactor.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -10,13 +10,21 @@ from src.constants import (
     STATIC_EATERIES_URL,
     PAY_METHODS,
     UPDATE_DELAY,
-    WEEKDAYS
+    WEEKDAYS,
 )
-from src import schema
+from src.schema import Data
+from src.types import (
+    CampusAreaType,
+    CoordinatesType,
+    EateryType,
+    EventType,
+    FoodItemType,
+    FoodStationType,
+    OperatingHoursType,
+    PaymentMethodsType,
+)
 
 eateries = {}
-events = {}
-menus = {}
 static_eateries = {}
 
 today = date.today()
@@ -29,11 +37,7 @@ def start_update():
     parse_eatery(data_json)
     statics_json = get(STATIC_EATERIES_URL).json()
     parse_static_eateries(statics_json)
-    schema.Data.update_data(
-        eateries=eateries,
-        events=events,
-        menus=menus,
-    )
+    Data.update_data(eateries)
   except Exception as e:
     print('Data update failed:', e)
   finally:
@@ -46,7 +50,7 @@ def parse_eatery(data_json):
     phone = eatery.get('contactPhone', 'N/A')
     phone = phone if phone else 'N/A'  # handle None values
 
-    new_eatery = schema.EateryType(
+    new_eatery = EateryType(
         about=eatery.get('about', ''),
         about_short=eatery.get('aboutshort', ''),
         campus_area=parse_campus_area(eatery),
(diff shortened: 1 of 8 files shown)


young-k:
Rebased & addressed linting comments

src/data.py
```diff
@@ -46,7 +50,7 @@ def parse_eatery(data_json):
     phone = eatery.get('contactPhone', 'N/A')
     phone = phone if phone else 'N/A'  # handle None values
```
kq-li, line 51: `phone or 'N/A'` since `phone` shouldn't be falsy anyway

kq-li: not your code but just something i noticed

src/types/account_info_type.py
```diff
@@ -0,0 +1,12 @@
+from graphene import List, String, ObjectType
+
+class TransactionType(ObjectType):
+  name = String(required=True)
+  timestamp = String(required=True)
+
+class AccountInfoType(ObjectType):
+  brbs = String(required=True)
+  cityBucks = String(required=True)
```
kq-li: `city_bucks`

src/schema.py
```diff
@@ -1,107 +1,36 @@
...
-  timestamp = String(required=True)
-
-class AccountInfoType(ObjectType):
-  brbs = String(required=True)
-  cityBucks = String(required=True)
-  history = List(TransactionType, required=True)
-  laundry = String(required=True)
-  swipes = String(required=True)
+  def update_data(eateries):
+    Data.eateries = eateries
```
kq-li, line 21: no more events or menus?

kq-li (commented):
good changes -- a few comments

src/schema.py
```diff
@@ -1,107 +1,36 @@
 from datetime import datetime
 from dateutil import parser
 
-from graphene import Field, ObjectType, String, List, Int, Float, Boolean
-from graphene.types.datetime import Date, Time
+from graphene import Boolean, Field, Int, List, String, ObjectType
```
meganle: alphabetize

src/types/account_info_type.py
```diff
@@ -0,0 +1,12 @@
+from graphene import List, String, ObjectType
```
meganle: same here

src/types/eatery_type.py
```diff
@@ -0,0 +1,28 @@
+from graphene import Field, Float, Int, List, ObjectType, String
```
meganle, line 1: same

src/schema.py
henrytso: I think events (`EventType`) got moved to `types/operating_hours_type.py` and menus (`List(FoodStationType)`) got moved to `types/food_station_type.py`.

henrytso (commented):
Looks much cleaner

src/schema.py
young-k: they're never queried, everything is from `eateries`

src/types/account_info_type.py
young-k: 😠 i didn't write this, but i'll make the change

src/data.py
young-k: phone can be `None` if in the dictionary `contactPhone: None`

kq-li: yeah i meant falsy but not `None`, the `or` will handle `None`s properly so it's just slightly cleaner

young-k:
@kq-li @meganle @henrytso addressed comments

kq-li (approved):
lgtm

PR #38, opened 2018-11-13
Fix bug regarding null laundry value

young-k:
Verified that it works, I don't have a laundry plan:

screen shot 2018-11-13 at 2 40 01 pm


kq-li:
weird that it's a string but lgtm

kq-li (approved):
weird that it's a string but lgtm

PR #42, opened 2018-11-15
Add Trillium menu from DiningStack

ilikesymmetry:
Access hardcodedMenus.json from DiningStack.  Sets menu for trillium in graphql while looping through cornell dining's json.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -9,6 +9,7 @@ CORNELL_DINING_URL = 'https://now.dining.cornell.edu/api/1.0/dining/eateries.jso
 CORNELL_INSTITUTION_ID = '73116ae4-22ad-4c71-8ffd-11ba015407b1'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
 GIT_CONTENT_URL = 'https://raw.githubusercontent.com/cuappdev'
+STATIC_MENUS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/hardcodedMenus.json'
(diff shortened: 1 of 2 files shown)


src/constants.py
```diff
@@ -9,6 +9,7 @@
 CORNELL_INSTITUTION_ID = '73116ae4-22ad-4c71-8ffd-11ba015407b1'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
 GIT_CONTENT_URL = 'https://raw.githubusercontent.com/cuappdev'
+STATIC_MENUS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/hardcodedMenus.json'
```
young-k: u already know, alphabetize

young-k (approved):
nifty nifty

src/data.py
```diff
@@ -253,3 +258,7 @@ def format_time(start_time, end_time, start_date):
   new_end = "{}:{}:{}".format(end_date, end_hour, end_minute)
 
   return [new_start, new_end]
+
+def trillium_menu():
```
young-k: can we call this `get_trillium_menu` instead?

src/constants.py
ilikesymmetry: AGHHHH

src/data.py
young-k: @cswenberg this 1 too

ilikesymmetry:
@young-k

young-k (approved):
beautiful

PR #44, opened 2018-11-15
Add brb and city bucks default values

ilikesymmetry:


src/schema.py
```diff
@@ -72,7 +72,9 @@ def resolve_account_info(self, info, session_id=None):
         }
     ).json()['response']['accounts']
 
-    account_info['laundry'] = "0.00"  # initialize default
+    account_info['brbs'] = '0.00'  # initialize default
+    account_info['city_bucks'] = '0.00'  # initialize default
+    account_info['laundry'] = '0.00'  # initialize default
```
young-k: let's get rid of `initialize default`, and just put a comment above the first one:
`initialize defaults`

young-k (approved):
Rebase the commit using `git add -A; git commit --amend`, call it "Add defaults to brbs and city_bucks" instead

PR #46, opened 2018-11-18
Add Collegetown eateries from Yelp

ilikesymmetry:
Use yelp's api to grab information from all c-town restaurants.  Manipulate the yelp data to fit GraphQL, added new field in schema and query route to get just collegetown eateries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,4 @@ MANIFEST
 
 # Environments
 .env
-.venv
 venv/
-venv2/
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ MANIFEST
 .env
 .venv
 venv/
+venv2/
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,3 +25,4 @@ six==1.11.0
 typing==3.6.6
 urllib3==1.24.1
 Werkzeug==0.14.1
+yelpapi==2.3.0
--- a/src/collegetown.py
+++ b/src/collegetown.py
@@ -1,48 +1,33 @@
 from os import environ
 from time import sleep
 
-from pprint import pprint
 from yelpapi import YelpAPI
 
 from src.constants import (
     YELP_LATITUDE,
     YELP_LONGITUDE,
     YELP_RADIUS,
-    YELP_NUMBER_LIMIT,
+    YELP_RESTAURANT_LIMIT,
     YELP_QUERY_DELAY,
 )
 
-API_KEY = environ.get('YELP_API_KEY')
-
-yelp_api = YelpAPI(API_KEY)
-
-def yelp_search(text, latitude=YELP_LATITUDE, longitude=YELP_LONGITUDE):
-  try:
-    print('searching..')
-    search = yelp_api.search_query(term=text, longitude=longitude, latitude=latitude, limit=5)
-    eatery = search['businesses'][0]
-    eatery_id = eatery['id']
-    id_query = yelp_api.business_query(id=eatery_id)
-    pprint(id_query)
-  except Exception as e:
-    print('something went wrong: ' + e)
+yelp_api = YelpAPI(environ.get('YELP_API_KEY'))
 
 def collegetown_search():
   try:
     search = yelp_api.search_query(
-        longitude=YELP_LONGITUDE,
         latitude=YELP_LATITUDE,
-        radius=YELP_RADIUS,
-        limit=YELP_NUMBER_LIMIT
+        limit=YELP_RESTAURANT_LIMIT,
+        longitude=YELP_LONGITUDE,
+        radius=YELP_RADIUS
     )
-    # print(len(search['businesses']))  #yelp will return at most 50
     eateries = []
     for place in search['businesses']:
       eatery_id = place['id']
       id_query = yelp_api.business_query(id=eatery_id)
-      # pprint(id_query)
       eateries.append(id_query)
       sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
+    print('done with collegetown search')
     return eateries
   except Exception as e:
     print(e)
--- a/src/collegetown.py
+++ b/src/collegetown.py
@@ -6,9 +6,9 @@ from yelpapi import YelpAPI
 from src.constants import (
     YELP_LATITUDE,
     YELP_LONGITUDE,
+    YELP_QUERY_DELAY,
     YELP_RADIUS,
     YELP_RESTAURANT_LIMIT,
-    YELP_QUERY_DELAY,
 )
 
 yelp_api = YelpAPI(environ.get('YELP_API_KEY'))
@@ -27,7 +27,7 @@ def collegetown_search():
       id_query = yelp_api.business_query(id=eatery_id)
       eateries.append(id_query)
       sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
-    print('done with collegetown search')
+    print('Done with Collegetown Search')
     return eateries
   except Exception as e:
     print(e)
--- a/src/collegetown.py
+++ b/src/collegetown.py
@@ -1,0 +1,48 @@
+from os import environ
+from time import sleep
+
+from pprint import pprint
+from yelpapi import YelpAPI
+
+from src.constants import (
+    YELP_LATITUDE,
+    YELP_LONGITUDE,
+    YELP_RADIUS,
+    YELP_NUMBER_LIMIT,
+    YELP_QUERY_DELAY,
+)
+
+API_KEY = environ.get('YELP_API_KEY')
+
+yelp_api = YelpAPI(API_KEY)
+
+def yelp_search(text, latitude=YELP_LATITUDE, longitude=YELP_LONGITUDE):
+  try:
+    print('searching..')
+    search = yelp_api.search_query(term=text, longitude=longitude, latitude=latitude, limit=5)
+    eatery = search['businesses'][0]
+    eatery_id = eatery['id']
+    id_query = yelp_api.business_query(id=eatery_id)
+    pprint(id_query)
+  except Exception as e:
+    print('something went wrong: ' + e)
+
+def collegetown_search():
+  try:
+    search = yelp_api.search_query(
+        longitude=YELP_LONGITUDE,
+        latitude=YELP_LATITUDE,
+        radius=YELP_RADIUS,
+        limit=YELP_NUMBER_LIMIT
+    )
+    # print(len(search['businesses']))  #yelp will return at most 50
+    eateries = []
+    for place in search['businesses']:
+      eatery_id = place['id']
+      id_query = yelp_api.business_query(id=eatery_id)
+      # pprint(id_query)
+      eateries.append(id_query)
+      sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
+    return eateries
+  except Exception as e:
+    print(e)
--- a/src/constants.py
+++ b/src/constants.py
@@ -34,3 +34,8 @@ WEEKDAYS = {
     'saturday': [5],
     'sunday': [6]
 }
+YELP_LATITUDE = 42.440680
+YELP_LONGITUDE = -76.486043
+YELP_NUMBER_LIMIT = 50
+YELP_RADIUS = 200  # units meters
+YELP_QUERY_DELAY = .1  # units seconds
--- a/src/constants.py
+++ b/src/constants.py
@@ -36,6 +36,6 @@ WEEKDAYS = {
 }
 YELP_LATITUDE = 42.440680
 YELP_LONGITUDE = -76.486043
-YELP_NUMBER_LIMIT = 50
-YELP_RADIUS = 200  # units meters
-YELP_QUERY_DELAY = .1  # units seconds
+YELP_RADIUS = 200  # meters
+YELP_RESTAURANT_LIMIT = 50  # maximum restaurants returned per query (as allowed by yelp)
+YELP_QUERY_DELAY = .3  # seconds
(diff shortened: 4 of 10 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ MANIFEST
 .env
 .venv
 venv/
+venv2/
```
young-k: why do you need a `venv2/`?

.gitignore
```diff
@@ -22,3 +22,4 @@ MANIFEST
 .env
 .venv
```
young-k: why do you need `.venv`?

requirements.txt
```diff
@@ -25,3 +25,4 @@ six==1.11.0
 typing==3.6.6
 urllib3==1.24.1
 Werkzeug==0.14.1
+yelpapi==2.3.0
```
young-k, line 28: completely completely unrelated, but this looks like `yell papi`

src/collegetown.py
```diff
@@ -0,0 +1,35 @@
...
+from time import sleep
+
+from yelpapi import YelpAPI
+
+from src.constants import (
+    YELP_LATITUDE,
+    YELP_LONGITUDE,
+    YELP_RADIUS,
+    YELP_NUMBER_LIMIT,
+    YELP_QUERY_DELAY,
```
young-k, line 9: this is alphabetized, impressive

src/collegetown.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+API_KEY = environ.get('YELP_API_KEY')
+
+yelp_api = YelpAPI(API_KEY)
```
young-k: why not just initialize it with `environ.get()` here, if it's not used anywhere else?

src/collegetown.py
```diff
@@ -0,0 +1,35 @@
...
+
+def collegetown_search():
+  try:
+    search = yelp_api.search_query(
+        longitude=YELP_LONGITUDE,
+        latitude=YELP_LATITUDE,
+        radius=YELP_RADIUS,
+        limit=YELP_NUMBER_LIMIT
```
young-k: alphabetize

src/collegetown.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+    eateries = []
+    for place in search['businesses']:
+      eatery_id = place['id']
+      id_query = yelp_api.business_query(id=eatery_id)
+      eateries.append(id_query)
+      sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
```
young-k, line 29: clever, but you should never call `sleep` in a function because then you won't be able to handle more client requests, is there a better way to do this?

src/constants.py
```diff
@@ -34,3 +34,8 @@
     'saturday': [5],
     'sunday': [6]
 }
+YELP_LATITUDE = 42.440680
+YELP_LONGITUDE = -76.486043
+YELP_NUMBER_LIMIT = 50  # maximum allowed by yelp
```
young-k: maximum what allowed by yelp? i don't think you need `number`, `limit` implies it's a number

src/constants.py
```diff
@@ -34,3 +34,8 @@
...
+YELP_RADIUS = 200  # units meters
+YELP_QUERY_DELAY = .3  # units seconds
```
young-k: you can just put seconds, and not `units`, same with the above

src/data.py
```diff
@@ -241,19 +246,34 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
 
-def format_time(start_time, end_time, start_date):
-  start_hour, start_minute = start_time.split(':')
-  end_hour, end_minute = end_time.split(':')
+def format_time(start_time, end_time, start_date, hr24=False, overnight=False):
+  if hr24:
+    start_hour = int(start_time[0:2])
+    start_minute = start_time[2:] + 'am' if start_hour < 12 else start_time[2:] + 'pm'
+    end_hour = int(end_time[0:2])
+    end_minute = end_time[2:] + 'am' if end_hour < 12 else end_time[2:] + 'pm'
```
young-k: there has to be a better way to parse 24 hour time into what you want, just Stack Overflow it and look for a clean solution

young-k (changes requested):
Minor changes requested as always

.gitignore
kq-li: `venv` ignoring should go in your global gitignore

src/data.py
```diff
@@ -3,6 +3,7 @@
 
 from requests import get
```
kq-li: i think i'd actually prefer using `requests.get` just because `get` is so ubiquitous that without the prefix it can be confusing, old code i know but worth a second look

src/data.py
```diff
@@ -262,3 +282,72 @@ def format_time(start_time, end_time, start_date):
...
+    new_eatery = EateryType(
+        about='',
+        about_short='',
+        campus_area=CampusAreaType(
+            description='Collegetown',
+            description_short='C-town'
+        ),
+        coordinates=parse_coordinates(eatery),
+        collegetown=True,
+        eatery_type='Collegetown Resturant',
```
kq-li: `Resturant` => `Restaurant`

src/schema.py
```diff
@@ -24,19 +24,18 @@ class Query(ObjectType):
...
       AccountInfoType,
       date=Date(),
       session_id=String(name='id')
   )
 
-  def resolve_eateries(self, info, eatery_id=None):
-    if eatery_id is None:
+  def resolve_eateries(self, info, eatery_id=None, collegetown=False):
+    if collegetown:
+      return [eatery for eatery in Data.eateries.values() if eatery.collegetown]
```
kq-li: may want to change this to `eatery.external` or something for future proofing

src/types/eatery_type.py
```diff
@@ -16,6 +16,7 @@ class EateryType(ObjectType):
   about_short = String(required=True)
   campus_area = Field(CampusAreaType, required=True)
   coordinates = Field(CoordinatesType, required=True)
+  collegetown = Boolean(required=True)
```
kq-li: think this is a reasonable use case for a non required variable, confirm with ios maybe

kq-li (commented):
this is dope as hell i'm really looking forward to using this one, just some minor thoughts

src/collegetown.py
ilikesymmetry: Clients won't be calling this function at all.  Only the start_update() that gets called once a day

src/collegetown.py
ilikesymmetry: I don't get what you mean

ilikesymmetry: ohh jk lol

.gitignore
young-k: nah i think it's fine, but definitely not a `venv2`

kq-li: naming it `venv` is def a dev choice though, other devs shouldn't have to name their environments the same as yours or have to add them all to ignore

src/collegetown.py
young-k: you unalphabetized it, impressive @cswenberg

src/constants.py
```diff
@@ -36,6 +36,6 @@
 }
 YELP_LATITUDE = 42.440680
 YELP_LONGITUDE = -76.486043
-YELP_NUMBER_LIMIT = 50  # maximum allowed by yelp
-YELP_RADIUS = 200  # units meters
-YELP_QUERY_DELAY = .3  # units seconds
+YELP_LIMIT = 50  # maximum restaurants returned per query (as allowed by yelp)
```
young-k: maybe rename to `YELP_RESTAURANT_LIMIT`, and alphabetize?

src/data.py
```diff
@@ -248,39 +248,30 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
-    elif start_hour == 0:
-      start_hour = 12
-    if end_hour > 13:
-      end_hour = end_hour - 12
-    elif end_hour == 0:
-      end_hour = 12
+    start = datetime.strptime(start_time, '%H%M')
+    s_time = start.strftime('%I:%M%p')
+    end = datetime.strptime(end_time, '%H%M')
+    e_time = end.strftime('%I:%M%p')
```
young-k: not a fan of having shorter variable names, keep it `start_time` and `end_time`

src/collegetown.py
```diff
@@ -0,0 +1,33 @@
...
+from time import sleep
+
+from yelpapi import YelpAPI
+
+from src.constants import (
+    YELP_LATITUDE,
+    YELP_LONGITUDE,
+    YELP_RADIUS,
+    YELP_RESTAURANT_LIMIT,
+    YELP_QUERY_DELAY,
```
young-k: unfortunate but `Q` is before `R`

src/collegetown.py
```diff
@@ -0,0 +1,33 @@
...
+        longitude=YELP_LONGITUDE,
+        radius=YELP_RADIUS
+    )
+    eateries = []
+    for place in search['businesses']:
+      eatery_id = place['id']
+      id_query = yelp_api.business_query(id=eatery_id)
+      eateries.append(id_query)
+      sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
+    print('done with collegetown search')
```
young-k: `Done with Collegetown Search`

src/constants.py
```diff
@@ -34,3 +34,8 @@
     'saturday': [5],
     'sunday': [6]
 }
+YELP_LATITUDE = 42.440680
+YELP_LONGITUDE = -76.486043
+YELP_RADIUS = 200  # meters
+YELP_RESTAURANT_LIMIT = 50  # maximum restaurants returned per query (as allowed by yelp)
+YELP_QUERY_DELAY = .4  # seconds
```
young-k: always better to have `0.4`

src/data.py
```diff
@@ -241,24 +248,90 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+        id=new_id,
+        image_url=eatery['image_url'],
+        name=eatery.get('name', ''),
+        operating_hours=parse_collegetown_hours(eatery),
+        payment_methods=PaymentMethodsType(
+            brbs=False,
+            cash=True,
+            cornell_card=False,
+            swipes=False,
+            credit=True,
```
young-k, line 299: alphabetize these

src/data.py
```diff
@@ -241,24 +248,90 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+  for eatery in collegetown_data:
+    new_id = resolve_id(eatery, collegetown=True)
+    new_eatery = CollegetownEateryType(
+        address=eatery['location']['address1'],
+        categories=[cuisine['title'] for cuisine in eatery.get('categories', [])],
+        coordinates=parse_coordinates(eatery),
+        eatery_type='Collegetown Restaurant',
+        id=new_id,
+        image_url=eatery['image_url'],
+        name=eatery.get('name', ''),
```
young-k, line 293: is this field actually nullable?

src/data.py
```diff
@@ -241,24 +248,90 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+            cash=True,
+            cornell_card=False,
+            swipes=False,
+            credit=True,
+            mobile=False,
+        ),
+        phone=eatery.get('phone', 'N/A'),
+        price=eatery.get('price', ''),
+        url=eatery.get('url', ''),
+
```
young-k, line 306: delete

src/types/eatery_type.py
```diff
@@ -12,17 +12,26 @@ class CoordinatesType(ObjectType):
...
-  location = String(required=True)
   name = String(required=True)
-  name_short = String(required=True)
-  operating_hours = List(OperatingHoursType, required=True)
   payment_methods = Field(PaymentMethodsType, required=True)
   phone = String(required=True)
+
+class CampusEateryType(EateryType):
+  about = String(required=True)
+  about_short = String(required=True)
```
young-k, line 25: i are both of these actually used by client? if not, can we keep the one that's actually used

src/schema.py
```diff
@@ -1,44 +1,54 @@
...
   )
 
-  def resolve_eateries(self, info, eatery_id=None):
+  def resolve_campus_eateries(self, info, eatery_id=None):
+    if eatery_id is None:
+      return [eatery for eatery in Data.campus_eateries.values()]
+    eatery = Data.campus_eateries.get(eatery_id)
+    return [eatery] if eatery is not None else []
+
+  def resolve_collegetown_eateries(self, info, eatery_id=None):
```
young-k, line 51: not a big fan of the duplicated code between this and the method above, can we make a helper function that takes `eatery_id` and `eateries` and does the same thing –– unless you can see a cleaner way to not duplicate this code

src/types/eatery_type.py
```diff
@@ -12,17 +12,26 @@ class CoordinatesType(ObjectType):
...
+  campus_area = Field(CampusAreaType, required=True)
+  location = String(required=True)
+  name_short = String(required=True)
```
young-k, line 28: same thing regarding `name_short`, i understand that the `json` may have it but if it's unused we should toss it

src/data.py
```diff
@@ -241,24 +248,90 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+    )
+    collegetown_eateries[new_eatery.id] = new_eatery
+
+def parse_collegetown_hours(eatery):
+  hours_list = eatery.get('hours', [{}])[0].get('open', [])
```
young-k, line 311: these needs a comment explaining what it's doing

src/data.py
```diff
@@ -241,24 +248,90 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
     )
   return new_operating_hours
 
-def format_time(start_time, end_time, start_date):
-  start_hour, start_minute = start_time.split(':')
-  end_hour, end_minute = end_time.split(':')
+def format_time(start_time, end_time, start_date, hr24=False, overnight=False):
```
young-k, line 251: can we get a docstring explaining what this does

young-k (approved):
LGTM, squash some commits before you merge in, contact either kenneth/megan/daniel on Slack to figure out how to rebase –– Stuff like `Make hound happy` should always be squashed, I think this entire PR should probably be 2-3 commits max, so squash a bunch of them and you can `reword` some of the older commits to be more verbose.

PR #47, opened 2018-11-27
Add Collegetown Eateries

ilikesymmetry:
Created new subclass structure for eateries along with CollegetownHours and CollegetownEvent types.  Increased timeout delay from 30 seconds (default) to 60 seconds to prevent timeout on yelp queries.  Added format_time functionality to handle 24hr input formats.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -3,6 +3,7 @@ from threading import Timer
 
 from requests import get
 
+from src.collegetown import collegetown_search
 from src.constants import (
     CORNELL_DINING_URL,
     IMAGES_URL,
@@ -39,6 +40,8 @@ def start_update():
     parse_eatery(data_json)
     statics_json = get(STATIC_EATERIES_URL).json()
     parse_static_eateries(statics_json)
+    collegetown_eateries = collegetown_search()
+    parse_collegetown_eateries(collegetown_eateries)
     Data.update_data(eateries)
   except Exception as e:
     print('Data update failed:', e)
@@ -262,3 +265,6 @@ def format_time(start_time, end_time, start_date):
 def get_trillium_menu():
   statics_json = get(STATIC_MENUS_URL).json()
   return parse_dining_items(statics_json['Trillium'][0])
+
+def parse_collegetown_eateries(eateries):
+  return []
--- a/src/data.py
+++ b/src/data.py
@@ -1,7 +1,7 @@
 from datetime import date, datetime, timedelta
 from threading import Timer
 
-from requests import get
+import requests
 
 from src.collegetown import collegetown_search
 from src.constants import (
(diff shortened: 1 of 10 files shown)


src/data.py
```diff
@@ -241,24 +248,95 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+        event.get('start', ''),
+        event.get('end', ''),
+        event_date,
+        hr24=True,
+        overnight=event.get('is_overnight', False)
+    )
+    new_event = CollegetownEventType(
+        description='General',
+        end_time=end,
+        start_time=start,
```
kq-li, line 339: drop trailing comma

src/schema.py
```diff
@@ -1,46 +1,57 @@
...
+  def update_data(campus_eateries, collegetown_eateries):
+    Data.campus_eateries = campus_eateries
+    Data.collegetown_eateries = collegetown_eateries
 
 class Query(ObjectType):
-  eateries = List(
-      EateryType,
+  campus_eateries = List(
+      CampusEateryType,
+      eatery_id=Int(name='id'),
```
kq-li, line 30: here too

kq-li: and everywhere else

src/schema.py
```diff
@@ -1,46 +1,57 @@
...
       date=Date(),
       session_id=String(name='id')
   )
 
-  def resolve_eateries(self, info, eatery_id=None):
+  def get_eateries(eateries, eatery_id):
     if eatery_id is None:
-      return [eatery for eatery in Data.eateries.values()]
-    eatery = Data.eateries.get(eatery_id)
+      return [eatery for eatery in eateries.values()]
```
kq-li: `list(eateries.values())` should work

src/types/operating_hours_type.py
```diff
@@ -2,6 +2,15 @@
 
 from src.types.food_station_type import FoodStationType
 
+class CollegetownEventType(ObjectType):
```
kq-li, line 5: what's the reasoning for this change?

start_server.sh
```diff
@@ -1 +1 @@
-gunicorn -w 4 -b 0.0.0.0:5000 app:app
+gunicorn -w 4 -t 60 -b 0.0.0.0:5000 app:app
```
kq-li, line 1: what's this for?

kq-li (commented):
looks good, few comments

src/data.py
young-k: trailing commas are fine – also they're mandatory in JS so we might as well keep the convention

src/schema.py
young-k: nein

src/schema.py
young-k: ah very clever, can we make this change in a PR -> master @cswenberg?

young-k:
@kq-li this change is from `master` to `release`, just a heads up

start_server.sh
ilikesymmetry: peep the pr comment. its to prevent timeout when we query yelp

ilikesymmetry: because when we have 4 things pinging yelp with the same api key, we burn through our queryies/second 4x as fast so we reduce the ping rate accordingly.

src/types/operating_hours_type.py
ilikesymmetry: we were just putting in empty strings for most of the attributes in the main eatery event type, since I made a new ctown eatery type, might as well make types for the hours/events too

src/schema.py
ilikesymmetry: on it

src/data.py
kq-li: well i guess we should just do one or the other instead of sometimes doing it then (also not mandatory, only recommended by airbnb iirc)

kq-li:
ah yes good point

meganle (approved):

PR #53, opened 2018-12-06
Fix empty menus from extended hours

ilikesymmetry:
When cornell dining extends some of the eatery hours (in particular for finals), they add a new time slot but do not repaste a new menu with it, leaving us with no dining items to show users.  We are default showing the menu for the earlier time slot.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -42,6 +42,7 @@ def start_update():
     dining_query = requests.get(CORNELL_DINING_URL)
     data_json = dining_query.json()
     parse_eatery(data_json)
+    fill_empty_menus(campus_eateries)
     statics_json = requests.get(STATIC_EATERIES_URL).json()
     parse_static_eateries(statics_json)
     yelp_query = collegetown_search()
@@ -127,6 +128,8 @@ def parse_events(event_list, eatery_id, event_date):
 
 def parse_food_stations(station_list, eatery_id):
   new_stations = []
+  if len(station_list) == 1 and not station_list[0]['items']:  # no menu actually provided
+    return new_stations
   for station in station_list:
     default_index = len(new_stations)
     station_items = parse_food_items(station['items'])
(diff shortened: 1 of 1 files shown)


src/data.py
```diff
@@ -127,6 +128,8 @@ def parse_events(event_list, eatery_id, event_date):
 
 def parse_food_stations(station_list, eatery_id):
   new_stations = []
+  if len(station_list) == 1 and len(station_list[0]['items']) == 0:  # no menu actually provided
```
young-k: i believe the `python-ic` way to write the right boolean is `and not station_list[0]['items']`: https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty

src/data.py
```diff
@@ -280,6 +283,19 @@ def get_trillium_menu():
   statics_json = requests.get(STATIC_MENUS_URL).json()
   return parse_dining_items(statics_json['Trillium'][0])
 
+def fill_empty_menus(eateries):
+  for eatery in eateries.values():
+    for operating_hour in eatery.operating_hours:
+      # need to choose event with data we want
+      if len(operating_hour.events) <= 1 or len(operating_hour.events[0].menu) == 0:  # ignore these
```
young-k: same comment regarding the right statement

src/data.py
```diff
@@ -280,6 +283,19 @@ def get_trillium_menu():
...
+        continue
+      base_event = operating_hour.events[0]
+      for event in operating_hour.events:
+        if len(event.menu) == 0:
```
young-k: same same

src/data.py
```diff
@@ -280,6 +283,19 @@ def get_trillium_menu():
...
+          event.menu = base_event.menu
+          event.station_count = base_event.station_count
+
+
```
young-k: delete?

young-k (commented):
Will leave for someone else to approve, also add @alanna-zhou on future PRs

src/data.py
alanna-zhou: why is this one `<= 1 `and the one with stations `== 1`?

src/data.py
```diff
@@ -280,6 +283,19 @@ def get_trillium_menu():
   statics_json = requests.get(STATIC_MENUS_URL).json()
   return parse_dining_items(statics_json['Trillium'][0])
 
+def fill_empty_menus(eateries):
+  for eatery in eateries.values():
+    for operating_hour in eatery.operating_hours:
+      # need to choose event with data we want
+      if len(operating_hour.events) <= 1 or len(operating_hour.events[0].menu) == 0:  # ignore these
+        continue
+      base_event = operating_hour.events[0]
```
kq-li, line 292: will the first event have this menu+station count info for sure?

kq-li (commented):
looks good, address all comments + i'll approve

src/data.py
kq-li: ah i guess you addressed this in the PR description so nvm

src/data.py
ilikesymmetry: the one on line 131 with ==1 is to prevent us from making an empty FoodStaionType object.  When we pull from the cornell dining json, if there is no menu, they return by defualt a menu of one station with empty items.  This adds unnecessary information so we are stopping the code from making a worthless object and instead just return an empty list for the menu.  This also makes it easier in the later part to determine if we need to copy a menu.

ilikesymmetry: This one with <=1 is there to skip over operating hours with only one or less events because there will be no reason for us to copy information into a later event.

meganle (approved):

PR #54, opened 2018-12-09
Future proof static hours parsing

ilikesymmetry:


src/data.py
```diff
@@ -227,7 +227,14 @@ def parse_eatery_type(eatery):
 def parse_static_op_hours(hours_list, eatery_id, dining_items):
   weekdays = {}
   for hours in hours_list:
-    for weekday in WEEKDAYS[hours['weekday']]:
+    if '-' in hours['weekday']:
```
alanna-zhou, line 230: why would there be hyphens in the hours?

src/constants.py
```diff
@@ -26,13 +26,13 @@
...
 # [0, 1, 2, 3, 4, 5, 6]
 WEEKDAYS = {
-    'monday-wednesday': [0, 1, 2],
-    'monday-thursday': [0, 1, 2, 3],
-    'monday-friday': [0, 1, 2, 3, 4],
-    'thursday': [3],
-    'friday': [4],
-    'saturday': [5],
-    'sunday': [6]
+    'monday': 0,
```
alanna-zhou, line 29: how come this got changed from monday-wednesday to monday?

alanna-zhou (approved):
looks okay to me just curious about some things

src/data.py
ilikesymmetry: in the externalEateries file, we represent days as 'monday-wednesday' or just 'wednesday' for example.

src/constants.py
ilikesymmetry: the previous method of having 'monday-...' correspond to a list of indexes would require an exhaustive list of day combinations to include all edge cases, having the days correspond to an index and then make the list in a method is more streamlined approach.

PR #57, opened 2018-12-11
Merge extended hours with previous event

ilikesymmetry:


meganle (approved):
Looks good! Might wanna get rid of the fill_empty_menus function later, but we can look into that later since this is urgent

src/data.py
```diff
@@ -299,9 +301,25 @@ def fill_empty_menus(eateries):
...
           event.station_count = base_event.station_count
 
+def merge_hours(eateries):
+  for eatery in eateries.values():
+    for operating_hour in eatery.operating_hours:
+      # need to choose event with data we want to copy
+      if len(operating_hour.events) <= 1:  # ignore these
+        continue
+      base_event = operating_hour.events[0]
+      for event in operating_hour.events[:]:
```
young-k: `[:]`? this doesn't do anything, did you mean `[1:]`?

ilikesymmetry: [:] copies the list so I don't mess up the for loop if I remove an item right?

young-k: hm nifty but difficult to read, can you leave a comment?

ilikesymmetry: https://unspecified.wordpress.com/2009/02/12/thou-shalt-not-modify-a-list-during-iteration/

PR #59, opened 2018-12-16
Add closing static eateries logic

ilikesymmetry:
This will allow us to make simple alterations to the dining stack externalEateries.json file to close the static eateries on specific dates and ranges.

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -204,8 +204,13 @@ def parse_static_eateries(static_json):
         location=eatery.get('location', ''),
         name=eatery.get('name', ''),
         name_short=eatery.get('nameshort', ''),
-        operating_hours=parse_static_op_hours(eatery['operatingHours'], new_id, dining_items),
-        payment_methods=parse_payment_methods(eatery['payMethods']),
+        operating_hours=parse_static_op_hours(
+            eatery.get('operatingHours', []),
+            new_id,
+            dining_items,
+            eatery.get('datesClosed', [])
+        ),
+        payment_methods=parse_payment_methods(eatery.get('payMethods', [])),
         phone=eatery.get('contactPhone', 'N/A'),
         slug=eatery.get('slug', '')
     )
@@ -230,7 +235,7 @@ def parse_eatery_type(eatery):
   except Exception:
     return 'Unknown'
 
-def parse_static_op_hours(hours_list, eatery_id, dining_items):
+def parse_static_op_hours(hours_list, eatery_id, dining_items, dates_closed):
   weekdays = {}
   for hours in hours_list:
     if '-' in hours['weekday']:
@@ -247,18 +252,30 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
   new_operating_hours = []
   for i in range(NUM_DAYS_STORED_IN_DB):
     new_date = today + timedelta(days=i)
-    new_events = weekdays.get(new_date.weekday(), [])
+    for dates in dates_closed:
+      if '-' not in dates:
+        closed_date = datetime.strptime(dates, '%m/%d/%y').date()
+        if new_date == closed_date:
+          break
+      else:
+        start_str, end_str = dates.split('-')
+        start_date = datetime.strptime(start_str, '%m/%d/%y').date()
+        end_date = datetime.strptime(end_str, '%m/%d/%y').date()
+        if start_date <= new_date <= end_date:
+          break
+    else:
+      # new_date is not present in dates_closed, we can add this date to the db
+      new_events = weekdays.get(new_date.weekday(), [])
+      for event in new_events:
+        event['menu'] = dining_items
 
-    for event in new_events:
-      event['menu'] = dining_items
-
-    new_operating_hours.append(
-        OperatingHoursType(
-            date=new_date.isoformat(),
-            events=parse_events(new_events, eatery_id, new_date.isoformat()),
-            status='EVENTS'
-        )
-    )
+      new_operating_hours.append(
+          OperatingHoursType(
+              date=new_date.isoformat(),
+              events=parse_events(new_events, eatery_id, new_date.isoformat()),
+              status='EVENTS'
+          )
+      )
   return new_operating_hours
 
 def format_time(start_time, end_time, start_date, hr24=False, overnight=False):


src/data.py
```diff
@@ -261,6 +268,22 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
     )
   return new_operating_hours
 
+def is_closed(current_date, dates_closed):
+  """Returns boolean indicating if the specified date is present in a list of closed dates.
+  Parameter current_date: a datetime.date object of the date to compare
+  Parameter dates_closed: a list of strings either in form 'mm/dd/yy' for a singular date or
+  'mm/dd/yy-mm/dd/yy' for a range of dates.
+  """
+  for dates in dates_closed:
```
young-k: In terms of writing functions this isn't really good practice –– you want your helper functions to work independently.

For example, we can think about a different way to spec this function:

```python
def is_date_in_range(date_to_compare, start_date, end_date):
  """ Returns True if [date_to_compare] is between [start_date] and [end_date]
  All params are datetime objects.
  """
  return date_to_compare >= start_date and date_to_compare <= end_date
```

Much cleaner, right? The obvious counterargument is that this isn't how your data is formatted: but there's only one usage of this function, and you can add a little bit of preprocessing, making the code more readable (and saving lines). For the case where `start_date == end_date`, you can just pass the same date for both arguments –– you want to have the `types` of your params consistent (in your function it varies between a `list` and a `string`).

In general, I think a majority of the functions in this file can have the same concepts applied, and it would make the codebase more readable for people in the future!

young-k (commented):
Leaving for someone else to approve, I added @meganle as a reviewer!

src/data.py
```diff
@@ -204,7 +204,12 @@ def parse_static_eateries(static_json):
         location=eatery.get('location', ''),
         name=eatery.get('name', ''),
         name_short=eatery.get('nameshort', ''),
-        operating_hours=parse_static_op_hours(eatery['operatingHours'], new_id, dining_items),
+        operating_hours=parse_static_op_hours(
+            eatery['operatingHours'],
```
alanna-zhou: jw is `operatingHours` camel case because it's a constant?

ilikesymmetry: its camel because it comes from a json file, which I think is camel standard

src/data.py
ilikesymmetry: Okay so you want less in the helper and then do the preprocessing in parse_static_op_hours()?

src/data.py
kq-li: best to use `.get` here for consistency i think

src/data.py
kq-li: +1 to young and also the more atomic and modular your functions are, the more useful they'll be if you ever need them again in the future
in general, refactoring more code out isn't necessarily the proper approach -- sometimes doing some preprocessing is actually easier to understand

src/data.py
```diff
@@ -261,6 +268,22 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+def is_closed(current_date, dates_closed):
+  """Returns boolean indicating if the specified date is present in a list of closed dates.
+  Parameter current_date: a datetime.date object of the date to compare
+  Parameter dates_closed: a list of strings either in form 'mm/dd/yy' for a singular date or
+  'mm/dd/yy-mm/dd/yy' for a range of dates.
+  """
+  for dates in dates_closed:
+    if '-' not in dates and current_date.strftime('%m/%d/%y') == dates:
+      return True
+    start_date, end_date = dates.split('-')
```
kq-li: generally i think it's best not to reassign a var like this, especially since it's changing types (from string to Date). python's lack of typed variables is probably one of the worst things about it and we should generally stay away from relying on it

src/data.py
```diff
@@ -204,7 +204,12 @@ def parse_static_eateries(static_json):
...
         name=eatery.get('name', ''),
         name_short=eatery.get('nameshort', ''),
-        operating_hours=parse_static_op_hours(eatery['operatingHours'], new_id, dining_items),
+        operating_hours=parse_static_op_hours(
+            eatery['operatingHours'],
+            new_id,
+            dining_items,
+            eatery.get('datesClosed', [])
+        ),
         payment_methods=parse_payment_methods(eatery['payMethods']),
```
kq-li: and here (ik it's not part of your pr but as good a time as any to change it)

kq-li (commented):
lgtm, address comments and we gucci

src/data.py
ilikesymmetry: yeet :)

src/data.py
```diff
@@ -247,20 +252,40 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
+              status='EVENTS'
+          )
+      )
   return new_operating_hours
 
+def is_date_in_range(date_to_compare, start_date, end_date):
+  """ Returns True if [date_to_compare] is between [start_date] and [end_date]
+  All params are datetime.date objects.
+  """
+  return date_to_compare >= start_date and date_to_compare <= end_date
```
kq-li: `start_date <= date_to_compare <= end_date`

src/data.py
```diff
@@ -247,20 +252,40 @@ def parse_static_op_hours(hours_list, eatery_id, dining_items):
...
   for i in range(NUM_DAYS_STORED_IN_DB):
     new_date = today + timedelta(days=i)
-    new_events = weekdays.get(new_date.weekday(), [])
 
-    for event in new_events:
-      event['menu'] = dining_items
+    for dates in dates_closed:
+      if '-' not in dates:
+        closed_date = datetime.strptime(dates, '%m/%d/%y').date()
+        if is_date_in_range(new_date, closed_date, closed_date):
```
kq-li: is this just `new_date == closed_date`?

kq-li (commented):
couple more things

src/data.py
ilikesymmetry: mmm yeaaah

ilikesymmetry: should I just drop the helper all together and replace both uses of the function with the appropriate boolean?

kq-li: yeah that would be good lol

ilikesymmetry: lol so basically just moved the whole initial helper I made into another function

kq-li (approved):
lgtm, good stuff

PR #60, opened 2019-01-23
Change get brb logic

ilikesymmetry:
GET changed the string that correlates with a students brb balance by adding the word 'Spring'.  Trying to make a change that will handle most potential future changes by looking for the presence of the brb keyword string.


meganle (approved):
As long as you tested it, it looks good to go!

PR #69, opened 2019-02-11
added swipes data

AAAstorga:
I added swipe functionality so if you have a meal plan with swipe data, it will show up in eatery

### DIFF
--- a/src/schema.py
+++ b/src/schema.py
@@ -11,6 +11,8 @@ from src.constants import (
     CORNELL_INSTITUTION_ID,
     GET_URL,
     SWIPE_PLANS,
+    IGNORE_LOCATIONS,
+    SWIPE_PLANS,
 )
 from src.types import (
     AccountInfoType,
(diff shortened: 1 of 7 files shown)


src/schema.py
```diff
@@ -100,16 +100,16 @@ def resolve_account_info(self, info, session_id=None):
     account_info['brbs'] = '0.00'
     account_info['city_bucks'] = '0.00'
     account_info['laundry'] = '0.00'
+    account_info['swipes'] = ''
```
young-k, line 105: shouldn't this be initialized to `0`?

src/schema.py
```diff
@@ -100,16 +100,16 @@ def resolve_account_info(self, info, session_id=None):
...
       if acct['accountDisplayName'] == ACCOUNT_NAMES['citybucks']:
         account_info['city_bucks'] = str("{0:.2f}".format(round(acct['balance'], 2)))
       elif acct['accountDisplayName'] == ACCOUNT_NAMES['laundry']:
         account_info['laundry'] = str("{0:.2f}".format(round(acct['balance'], 2)))
       elif ACCOUNT_NAMES['brbs'] in acct['accountDisplayName']:
         account_info['brbs'] = str("{0:.2f}".format(round(acct['balance'], 2)))
-      # Need more research to implement swipes:
-      # Each plan has a different accountDisplayName
-      account_info['swipes'] = ''
+      elif any(meal_swipe_name in acct['accountDisplayName'] for meal_swipe_name in ['Bear Basic', 'Bear Choice', 'Bear Traditional']):
```
young-k: why not define `['Bear Basic', 'Bear Choice', 'Bear Traditional']` in a constant up top?

also, i am not a fan of this nested one-line for loop, but we can verify with @meganle

young-k:
Your commit message should also be `Add swipes data`

src/schema.py
AAAstorga: No sir

src/schema.py
AAAstorga: Not sure. I don't really do python backend, was just trying to fix a feature

meganle: yeah, i think it's best if the array is a constant.

also is there any reason that you use `any(...)`?

AAAstorga: stackoverflow

ilikesymmetry:
@young-k @meganle @keivanshahida @sahamanish99

meganle (approved):

keivanshahida:
This looks good to me - pretty straightforward changes.

Omarrasheed (approved):
LGTM fam -def backend dev

PR #70, opened 2019-02-11
Remove unused eatery attributes

ilikesymmetry:
Many attributes were not actually being utilized by iOS queries, so they have no purpose in our codebase.  Also removed an old method that was not being used


meganle (approved):
Nice job! As long as you verified with iOS it looks good

sahamanish99:
Looks good to me!

manishsaha (approved):
Looks good to me!

PR #71, opened 2019-02-18
Trim non-brb purchases

ilikesymmetry:
Cleans up our history from GET to only show the brb related purchases (excludes laundry, deposits, other cornell internal things).  Also strips the register number from the location


orkosinha (commented):
looks good

orkosinha (approved):

manishsaha (approved):
Looks good to me

PR #73, opened 2019-03-05
Add documentation to data.py

orkosinha:
Address issue #66

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -37,15 +37,24 @@ static_eateries = {}
 today = date.today()
 
 def start_update():
+  """Starts the data retrieval process.
+
+  Begins the process of getting data for the campus_eateries dictionary,
+  from Cornell Dining and the static source, and collegetown_eateries dictionary,
+  from Yelp.
+  """
   try:
     print('[{}] Updating campus'.format(datetime.now()))
+    # Get data for on-campus, Cornell-owned dining Options
     dining_query = requests.get(CORNELL_DINING_URL)
     data_json = dining_query.json()
     parse_eatery(data_json)
     merge_hours(campus_eateries)
+    # Get data for on-campus, 3rd-party dining options
     static_json = requests.get(STATIC_EATERIES_URL).json()
     parse_static_eateries(static_json)
     Data.update_data(campus_eateries)
+    # Get data for Collegetown eateries
     print('[{}] Updating collegetown'.format(datetime.now()))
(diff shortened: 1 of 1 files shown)


ilikesymmetry (approved):
so fresh and so cleeeaaaan

young-k (changes requested):
Can these messages be reformatted according to: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

```
def example():
    """ This should be a one sentence summary.

    Longer summaries that span multiple lines 
    should go here

    Args:
         ...
"""
```

young-k (commented):
The branch name should also be `orko/documentation`, you can rename a branch with `git branch -m [BRANCH_NAME]`, but that's for next time!

src/data.py
```diff
@@ -93,12 +112,19 @@ def parse_payment_methods(methods):
...
+      dining_items (list): A list that holds a dictionary for the items an eatery serves and a flag
+       for healthy options
+  """
   new_operating_hours = []
   hours_list = eatery['operatingHours']
   for hours in hours_list:
     new_date = hours.get('date', '')
     hours_events = hours['events']
-
+    #merge the dining items from the json (for non-dining hall eateries) with the menu for an event
```
meganle, line 137: Let's add a space between the '#' and 'merge'!

meganle (approved):
Everything looks good, great job! Just address Young's comment about the formatting and we can merge

src/data.py
orkosinha: Added it back!

src/data.py
```diff
@@ -93,12 +120,22 @@ def parse_payment_methods(methods):
...
+      dining_items (list): A list that holds a dictionary for the items an eatery serves and a flag
+       for healthy options
+  """
   new_operating_hours = []
   hours_list = eatery['operatingHours']
   for hours in hours_list:
     new_date = hours.get('date', '')
     hours_events = hours['events']
 
+    #merge the dining items from the json (for non-dining hall eateries) with the menu for an event
```
young-k, line 137: this space is still missing lol

young-k (approved):
LGTM, can you squash the two commits into one?

src/data.py
orkosinha: oops

orkosinha: squashed and resolved!

PR #75, opened 2019-03-10
File restructure for eateries

orkosinha:
Address Issue #64

### DIFF
--- a/src/eatery/campus_eatery.py
+++ b/src/eatery/campus_eatery.py
@@ -1,0 +1,88 @@
+import requests
+
+from src.constants import (
+    STATIC_MENUS_URL,
+    TRILLIUM_ID
+)
+from src.types import (
+    CampusEateryType,
+    OperatingHoursType
+)
+from src.eatery.common_eatery import (
+    get_image_url,
+    parse_campus_area,
+    parse_coordinates,
+    parse_dining_items,
+    parse_eatery_type,
+    parse_events,
+    parse_payment_methods,
+    resolve_id
+)
+
+def parse_eatery(data_json, campus_eateries):
+  """Parses a Cornell Dining json dictionary.
+
+  Fills the campus_eateries dictionary with objects of type CampusEateryType.
+
+  Args:
+      data_json (dict): a valid dictionary from the Cornell Dining json
+      campus_eateries (dict): a dictionary to fill with campus eateries
+  """
+  for eatery in data_json['data']['eateries']:
+    eatery_id = eatery.get('id', resolve_id(eatery))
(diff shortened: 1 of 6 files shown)


src/eatery/campus_eatery.py
```diff
@@ -0,0 +1,89 @@
+import requests
+
+from src.constants import (
+    STATIC_MENUS_URL,
+    TRILLIUM_ID,
```
meganle: Trailing comma here but not on lines 9 and 19

meganle (approved):
LGTM! Just peep the one comment I made

src/eatery/campus_eatery.py
keivanshahida: Any way we could use a linter to deal with these kinds of stylistic requirements? @meganle

src/eatery/campus_eatery.py
```diff
@@ -0,0 +1,89 @@
...
+def get_trillium_menu():
+  """Gets the Trillium menu.
+
+  Returns the Trillium dining items (using parse_dining_items) from the static json source
+  for menus.
+  """
+  static_json = requests.get(STATIC_MENUS_URL).json()
+  return parse_dining_items(static_json['Trillium'][0])
+
+
```
keivanshahida: Single or double lines between function declarations? @meganle

src/eatery/common_eatery.py
```diff
@@ -0,0 +1,207 @@
...
+  """Generates a URL for an image.
+
+  Creates a string that represents a url pointing towards the eatery image.
+
+  Args:
+      slug (string): a slugged name of an eatery
+  """
+  return "{}{}.jpg".format(IMAGES_URL, slug)
+
+
```
keivanshahida: Single or double line breaks?

keivanshahida (changes requested):
See questions, but otherwise LGTM!

ilikesymmetry (approved):
dope dope dope dude.  Just make those couple changes :)

src/eatery/campus_eatery.py
meganle: Good catch, we use flake8. Let's see if we can add rule to the config to prohibit trailing commas so that Hound catches this stuff @cswenberg @orkosinha

src/eatery/campus_eatery.py
meganle: Let's keep things consistent so single lines seems to be the norm

src/eatery/common_eatery.py
meganle: same thing here as above

keivanshahida (approved):

keivanshahida:
Send it dude!

PR #77, opened 2019-03-14
Fix parsing of Cornell's dining JSON

meganle:
- We were getting errors since we indexed the first item of all the menus if the menu existed
- Menus can be empty lists now in Cornell's JSON
- The start_update function wouldn't update the data cause of this which resulted in no eateries/menus

Also this is a pretty large if statement and wasn't sure how to format it best

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -66,8 +66,10 @@ def merge_hours(eateries):
         continue
       base_event = operating_hour.events[0]
       for event in operating_hour.events[1:]:  # iterate over copy of list so we can safely remove
-        if event.start_time == base_event.end_time and \
-            (not event.menu or event.menu[0].equals(base_event.menu[0])):
+        if (event.start_time == base_event.end_time and
+            event.menu and
+            base_event.menu and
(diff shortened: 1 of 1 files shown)


src/data.py
```diff
@@ -67,7 +67,8 @@ def merge_hours(eateries):
       base_event = operating_hour.events[0]
       for event in operating_hour.events[1:]:  # iterate over copy of list so we can safely remove
         if event.start_time == base_event.end_time and \
-            (not event.menu or event.menu[0].equals(base_event.menu[0])):
+            (not event.menu or (len(event.menu) > 0 and len(base_event.menu) > 0
+                and event.menu[0].equals(base_event.menu[0]))):
```
young-k: why not:

```python
if (event.start_time == base_event.end_time and 
    event.menu and 
    base_event.menu and 
    event.menu[0].equals(base_event.menu[0])):
...
```

Doesn't this accomplish the same thing while being easier to read?

young-k (approved):
LGTM

ilikesymmetry (approved):
lgtm :)

PR #78, opened 2019-03-25
Fix failed yelp requests

ilikesymmetry:
If our yelp requests crashed, we would not return anything back and cause NoneType errors.  Returning an empty list fixes this issue, but a more thorough investigation into yelp will be continued (although I could not recreate the error on my machine, unfortunately).


meganle (approved):
nice let's get this merged into release ASAP

orkosinha (approved):
lgtm!

PR #80, opened 2019-03-26
Fix merging of events

meganle:
If we have two events like this where the second event's menu is empty, they don't actually get combined. I changed it so that if the second event **does** have a menu then it checks for equality. So if the second doesn't have a menu it will default to the previous event's one.

Screen Shot 2019-03-26 at 6 32 34 PM

### DIFF
--- a/src/data.py
+++ b/src/data.py
@@ -68,10 +68,11 @@ def merge_hours(eateries):
       base_event = operating_hour.events[0]
       for event in operating_hour.events[1:]:  # iterate over copy of list so we can safely remove
         if (event.start_time == base_event.end_time and
-            event.menu and
             base_event.menu and
-            event.menu[0].equals(base_event.menu[0])):
+            (not event.menu or
+            event.menu[0].equals(base_event.menu[0]))):
           base_event.end_time = event.end_time
           operating_hour.events.remove(event)
           print('merged events for {} on {}'.format(eatery.name, operating_hour.date))
-        base_event = event
+        else:
+          base_event = event


ilikesymmetry (commented):
why are we adding event.menu = base_event.menu ?

src/data.py
```diff
@@ -68,10 +68,11 @@ def merge_hours(eateries):
...
-            event.menu and
             base_event.menu and
-            event.menu[0].equals(base_event.menu[0])):
+            (not event.menu or
+            event.menu[0].equals(base_event.menu[0]))):
           base_event.end_time = event.end_time
+          event.menu = base_event.menu
           operating_hour.events.remove(event)
           print('merged events for {} on {}'.format(eatery.name, operating_hour.date))
         base_event = event
```
meganle: @cswenberg we set `event.menu = base_event.menu` since either they both have the same menu already, or `event` has no menu like rpcc's light dinner. in that case, we give `event` the same menu as `base_event`. we do this since we set `base_event = event` later (in line 78), and we want our `base_event` to carry on the same menu.

ilikesymmetry: Okay, I think we should put `base_event = event` in an else block or put a continue statement after the print instead.  They'll do the same thing but make more sense readability-wise I feel.

src/data.py
```diff
@@ -68,10 +68,11 @@ def merge_hours(eateries):
       base_event = operating_hour.events[0]
       for event in operating_hour.events[1:]:  # iterate over copy of list so we can safely remove
         if (event.start_time == base_event.end_time and
-            event.menu and
             base_event.menu and
-            event.menu[0].equals(base_event.menu[0])):
+            (not event.menu or
+            event.menu[0].equals(base_event.menu[0]))):
```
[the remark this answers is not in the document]
young-k: @cswenberg can we add this to pylint to avoid these in the future

young-k (approved):
lgtm, nice fix

ilikesymmetry (approved):
yeet 👍

PR #81, opened 2019-04-04
Change brb history timestamps and fix eatery names

ilikesymmetry:
timestamps for brb transactions now read as <weekday>, <shortened month> <day number> at <HH:MM> AM/PM.

GET also was giving cryptic dining hall names (ex: "Duffield", "Statler Macs", "RPME", etc.) that are changed to reflect the names the same as Cornell Dining Now.

### DIFF
--- a/src/schema.py
+++ b/src/schema.py
@@ -11,6 +11,7 @@ from src.constants import (
     CORNELL_INSTITUTION_ID,
     GET_URL,
     IGNORE_LOCATIONS,
+    LOCATION_NAMES,
     SWIPE_PLANS,
 )
 from src.types import (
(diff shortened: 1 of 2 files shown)


src/schema.py
```diff
@@ -141,12 +142,13 @@ def resolve_account_info(self, info, session_id=None):
         continue
       dt_utc = parser.parse(t['actualDate']).astimezone(pytz.timezone('UTC'))
```
young-k: Are we using `dt_utc`? If not, why not `pytz.timezone('EST')` and just set `dt_est` from the start? Also, prefer more verbose variable names like `datetime_est`

src/schema.py
```diff
@@ -141,12 +142,13 @@ def resolve_account_info(self, info, session_id=None):
...
       dt_est = dt_utc.astimezone(pytz.timezone('US/Eastern'))
+      location = t['locationName'][:t['locationName'].rfind(' ')]
```
young-k: `location = t['locationName'].rsplit(' ', 1)[0]` is probably cleaner, but check that the code works

src/schema.py
```diff
@@ -141,12 +142,13 @@ def resolve_account_info(self, info, session_id=None):
...
+      # removes the register numbers at the end of the string by taking the substring up until
+      # the last space (right before the number)
       new_transaction = {
           'amount': t['amount'],
```
young-k: prefer more verbose names than `t`

young-k (approved):
LGTM, left comments above!

src/schema.py
ilikesymmetry: ohh so fresh and so cleeeaaan

orkosinha (approved):
lgtm!

PR #83, opened 2019-04-07
Alter Libe Cafe name

ilikesymmetry:


young-k (approved):
LGTM, but are we getting these checked somehow (maybe through a Cornell Dining contact), or are we eyeballing all of this?

ilikesymmetry:
@young-k just eyeballing, the name change is to keep it consistent with the names currently used in the app

meganle:
we can prob cross check with whatever is on cornell dining’s website

orkosinha (approved):
lgtm, I think one potential way to cross check GET names with Cornell Dining is to implement some sort of fuzzy search, but given how inconsistent the names are, idt that'd work well

PR #85, opened 2019-04-10
Add Enums for PaymentMethods and Ratings 

orkosinha:
Create a new field in eatery and Collegetown eatery types for payment methods that uses an Enum, and a new field only for the Collegetown eatery type for ratings that also uses an Enum.

### DIFF
--- a/src/eatery/campus_eatery.py
+++ b/src/eatery/campus_eatery.py
@@ -16,6 +16,7 @@ from src.eatery.common_eatery import (
     parse_eatery_type,
     parse_events,
     parse_payment_methods,
+    parse_payment_methods_enum,
     resolve_id
 )
 
@@ -46,6 +47,7 @@ def parse_eatery(data_json, campus_eateries):
         name_short=eatery.get('nameshort', ''),
         operating_hours=parse_operating_hours(eatery, dining_items),
         payment_methods=parse_payment_methods(eatery['payMethods']),
+        payment_methods_enums=parse_payment_methods_enum(eatery.get('payMethods', [])),
         phone=phone,
         slug=eatery.get('slug', '')
     )
--- a/src/eatery/common_eatery.py
+++ b/src/eatery/common_eatery.py
@@ -10,6 +10,7 @@ from src.types import (
     EventType,
     FoodItemType,
     FoodStationType,
+    PaymentMethodsEnum,
     PaymentMethodsType
 )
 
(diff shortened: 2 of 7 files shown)


src/eatery/common_eatery.py
```diff
@@ -115,6 +117,25 @@ def parse_payment_methods(methods):
...
+  if (any(pay['descrshort'] == PAY_METHODS['brbs'] for pay in methods)):
+      payment_methods.append(PaymentMethodsEnum.BRB)
+  if (any(pay['descrshort'] == PAY_METHODS['credit'] for pay in methods)):
+      payment_methods.append(PaymentMethodsEnum.CASH)
+      payment_methods.append(PaymentMethodsEnum.CREDIT)
+  if (any(pay['descrshort'] == PAY_METHODS['c-card'] for pay in methods)):
+      payment_methods.append(PaymentMethodsEnum.MOBILE)
+  if (any(pay['descrshort'] == PAY_METHODS['swipes'] for pay in methods)):
+      payment_methods.append(PaymentMethodsEnum.SWIPES)
+  return payment_methods
```
young-k, line 140: probably cleaner to:

```
for method in methods:
  description = method['descrshort']
  if description == PAY_METHODS['brbs']:
    ...
```

young-k (changes requested):
Versioning models is **strongly discouraged in GraphQL**, refer to: https://graphql.org/learn/best-practices/#versioning

Can this be done by adding new fields instead?

src/eatery/common_eatery.py
orkosinha: That makes a lot more sense, although we already have two queries (campusEatery and eateries) that give the exact same information. Should the new field be part of one of them or both of them?

orkosinha:
> Versioning models is **strongly discouraged in GraphQL**, refer to: https://graphql.org/learn/best-practices/#versioning
> 
> Can this be done by adding new fields instead?

That makes a lot more sense, although we already have two queries (campusEatery and eateries) that give the exact same information. Should the new field be part of one of them or both of them?

src/types/eatery_type.py
```diff
@@ -10,13 +10,25 @@ class CoordinatesType(ObjectType):
...
+class RatingEnum(Enum):
+  ONE = 2
+  ONE_HALF = 3
+  TWO = 4
+  TWO_HALF = 5
+  THREE = 6
+  THREE_HALF = 7
+  FOUR = 8
+  FOUR_HALF = 9
+  FIVE = 10
```
young-k, line 23: is there a better name for these values? maybe add a comment to the class to make it explicit what these represent

src/types/payment_methods_type.py
```diff
@@ -1,9 +1,17 @@
-from graphene import Boolean, ObjectType
+from graphene import Boolean, Enum, ObjectType
+
+class PaymentMethodsEnum(Enum):
+  BRB = 0
+  CASH = 1
+  CORNELL_CARD = 2
+  CREDIT = 3
+  MOBILE = 4
+  SWIPES = 5
```
young-k, line 9: do Enums need corresponding integer values?

young-k (approved):
LGTM, left two last comments

src/eatery/campus_eatery.py
```diff
@@ -46,6 +47,7 @@ def parse_eatery(data_json, campus_eateries):
         name_short=eatery.get('nameshort', ''),
         operating_hours=parse_operating_hours(eatery, dining_items),
         payment_methods=parse_payment_methods(eatery['payMethods']),
+        payment_methods_enum=parse_payment_methods_enum(eatery.get('payMethods', [])),
```
young-k: maybe `payment_method_enums` instead?

src/types/eatery_type.py
orkosinha: I'm not really sure what to name them, but I'll add a comment for clarification

src/types/payment_methods_type.py
orkosinha: The Graphene documentation for Enums says it needs some constant value and it's just a wrapper for the standard Python enum which I think always needs some value (which can be a string with the corresponding name)

src/eatery/common_eatery.py
```diff
@@ -115,6 +116,29 @@ def parse_payment_methods(methods):
...
+
+  Args:
+      methods (json): a valid json segment for payments from Cornell Dining
+  """
+  payment_methods = []
+  for method in methods:
+    description = method['descrshort']
+    if (description == PAY_METHODS['brbs']):
+      payment_methods.append(PaymentMethodsEnum.BRB)
+    if (description == PAY_METHODS['credit']):
```
ilikesymmetry, line 131: these could be `else if` yes?

ilikesymmetry (approved):
lgtm :)

PR #86, opened 2019-04-11
Add graceful_timeout to gunicorn and fix resolve_id

ilikesymmetry:
resolve_id was causing crashes when updating static eateries because of an attempt to use a nonexistent key.  Fixing up this method should prevent crashes and the graceful_timout increase to match the current timeout setting should prevent an infinite loop of worker timeouts :)


meganle (approved):
Great job!!

young-k (approved):
tbh i have no idea how this fixes it, but i believe you

PR #87, opened 2019-04-11
Fix event merging, GET information, resolve_id method

ilikesymmetry:
Addresses event merging for events with empty menus, the GET location names and reformatting of timestamps, and a bug caused by updating static eateries in the resolve_id method.  Hopefully this will fix our yelp issues, one can only look to the debug gods on this one.


meganle (approved):
let's gooo!

PR #89, opened 2019-04-15
Use curated images for Collegetown eateries

orkosinha:
Follows the same format as the campus eateries for creating an image URL and uses a unique alias from Yelp to identify each Collegetown eatery.


ilikesymmetry (approved):
lgtm :)

meganle (approved):
lgtm!!

PR #91, opened 2019-04-16
Swipe data integration for Wait Times

ilikesymmetry:
Wait Times have finally made it!!
We do this by parsing a `.log` file and along with pandas and numpy, construct a massive table to store all of the swipe data with which we run calculations on.  We have two files that perpetuate in a new folder `eatery-data` which is tied via docker volume to another directory in our deployed server.  The first file is 'data.csv' which is simply a tabular form of the 'data.log' file so allow us to more easily work with.  The second file is 'timeblock-averages.csv' which holds the calculations for our wait time bounds and what ultimately provides the data for our new swipe_data attribute in our Eatery objects.  I did my best to document the blocks of code and add docstrings to my methods.  Please roast and destroy my pr :)


.gitignore
```diff
@@ -2,6 +2,7 @@
 __pycache__/
 *.py[cod]
 *$py.class
+.DS_Store
```
young-k: should be in a global gitignore, refer to: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

.gitignore
```diff
@@ -2,6 +2,7 @@
...
 
 # Distribution / packaging
 .Python
```
young-k: do we need this?

README.md
```diff
@@ -21,6 +21,7 @@ The required environment variables for this API are the following:
 
 ````bash
 export FLASK_ENV=development
+export YELP_API_KEY=<insert>
```
young-k: can we drop the `export` keyword, `autoenv` doesn't require it and we can copy this straight into a server's configuration if the `export` keyword is removed.

env.template
```diff
@@ -1 +1,2 @@
 export FLASK_ENV=development
+export YELP_API_KEY=<insert>
```
young-k: same thing here, also maybe just put `API_KEY` instead since we don't have the `<>` convention anywhere else?

src/constants.py
```diff
@@ -3,39 +3,57 @@
...
 CORNELL_INSTITUTION_ID = '73116ae4-22ad-4c71-8ffd-11ba015407b1'
+DINING_HALL = 'dining_hall'
+EATERY_DATA_PATH = './eatery-data/'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
 GIT_CONTENT_URL = 'https://raw.githubusercontent.com/cuappdev'
 IGNORE_LOCATIONS = ['BS-No Bill Workstation', 'Admin Workstation (B)']
 IMAGES_URL = GIT_CONTENT_URL + '/assets/master/eatery/eatery-images/'
+TRILLIUM = 'trillium'
+ISOLATE_SWIPES = ['date', 'session_type', 'weekday', 'location', 'start_time', 'end_time',
+                    DINING_HALL, BRB_ONLY, TRILLIUM]
```
young-k: if this is on multiple lines, it should be:
```python
ISOLATE_SWIPES = [
    'a',
    'b',
    ...
]
```

src/constants.py
```diff
@@ -3,39 +3,57 @@
...
+ISOLATE_DATE = ['session_type', 'weekday', 'location', 'start_time', 'end_time', 'swipes', 
+                    DINING_HALL, BRB_ONLY, TRILLIUM]
```
young-k: same comment here, also shouldn't this be a plural rather than `DATE`?

src/constants.py
```diff
@@ -3,39 +3,57 @@
...
+ISOLATE_COUNTER_SWIPES = ['session_type', 'weekday', 'location', 'start_time', 'end_time',
```
young-k: same comment

src/constants.py
```diff
@@ -45,14 +63,24 @@
...
 }
+SCHOOL_BREAKS = {
+  'fall': '10/6/18-10/9/18',
+  'thanksgiving': '11/21/18-11/25/18',
+  'finals_winter': '12/5/18-12/15/18',
+  'winter': '12/16/18-1/21/19',
+  'february': '2/23/19-2/26/19',
+  'spring': '3/30/19-4/7/19',
+  'finals_spring': '5/8/19-5/18/19',
+  'summer': '5/19/19-8/28/19'
```
young-k, line 80: not sure what this is used for, but shouldn't this be school year independent? else this should be added to documentation as well, that it needs to be updated yearly

src/data.py
```diff
@@ -15,9 +15,11 @@
     UPDATE_DELAY
 )
 from src.schema import Data
+from src.swipes import export_data, parse_to_csv
 
 campus_eateries = {}
 collegetown_eateries = {}
+all_swipe_data = {}
```
young-k: alphabetize

src/eatery/static_eatery.py
```diff
@@ -108,3 +107,12 @@ def parse_static_op_hours(hours_list, dining_items, dates_closed):
           )
       )
   return new_operating_hours
+
+def string_to_date_range(dates):
+  """
+  date: string representing a range of dates mm/dd/yy-mm/dd/yy
```
young-k: this is `date` but the param is `dates`?

src/schema.py
```diff
@@ -144,9 +144,10 @@ def resolve_account_info(self, info, session_id=None):
       location = txn['locationName'].rsplit(' ', 1)[0]
       # removes the register numbers at the end of the string by taking the substring up until
       # the last space (right before the number)
+      name = LOCATION_NAMES.get(location, {'name': location})['name']
       new_transaction = {
           'amount': txn['amount'],
-          'name': LOCATION_NAMES.get(location, location),
+          'name': name if name != 'Amit Bhatia Libe Café' else 'Libe Café',
```
young-k: do we really need additional parsing here? there isn't a way to put this in the previous snippet?

src/schema.py
```diff
@@ -144,9 +144,10 @@ def resolve_account_info(self, info, session_id=None):
       location = txn['locationName'].rsplit(' ', 1)[0]
       # removes the register numbers at the end of the string by taking the substring up until
       # the last space (right before the number)
+      name = LOCATION_NAMES.get(location, {'name': location})['name']
```
young-k: this code isn't intuitive, probably more readable to do:
```python
name = LOCATION_NAMES['location']['name'] if location in LOCATION_NAMES else location
```

Also, `LOCATION_NAMES` isn't a good name for this constant anymore!

src/swipes.py
```diff
@@ -0,0 +1,304 @@
+from os.path import isfile
+from datetime import date, datetime, timedelta
+import json
```
young-k, line 3: prefer to alphabetize the first three imports

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+  ISOLATE_SWIPES,
+  SCHOOL_BREAKS,
+  TRILLIUM,
+  WAIT_TIME_CONVERSION,
+  WEEKDAYS,
+)
+from src.eatery import string_to_date_range
+from src.types import SwipeDataType
+
+weekdays = {index: day for day, index in WEEKDAYS.items()}  # invert our weekday -> index dict
```
young-k: might be more readable & cleaner to do:

`{v:k for k, v in ...}`, just to make it EXPLICIT that the keys & values are being inverted, and the comment should explain why this is happening

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+breaks = {}
+for label, dates in SCHOOL_BREAKS.items():
+  breaks[label] = string_to_date_range(dates)
+
+def parse_to_csv(file_name='data.csv', limit=-1):
+  """
+  Takes in a data.log file of swipe logs and converts them to a tabular format.
```
young-k: can you follow https://www.python.org/dev/peps/pep-0257/

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+      df = pd.concat([update_info['new_marker'], df])
+      df.to_csv(MAIN_CSV, header=True, index=False)
+      return data_file
+  except Exception as e:
+    print('Failed at parse_to_csv')
+    print('Data update failed:', e)
+
+def sort_by_timeblock(input_file_path, output_file='timeblock-averages.csv'):
+  """
+  Sorts and runs average swipe/time calculations.
```
young-k: same thing here: https://www.python.org/dev/peps/pep-0257/

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    return output_path
+  except Exception as e:
+    print('Failed at sort_by_timeblock')
+    print('Data update failed:', e)
+
+
+def sort_by_day(input_file_path, output_file='daily-averages.csv'):
+  """
+  Sorts and runs average swipe/time calculations.
+  Saves to a csv file.  Currently not used in main code, but allows for future use of data
```
young-k: ok last time regarding docstrings, fix it everywhere though!

young-k (changes requested):
Going to leave more comments, you might want to get started on these first since reading the code will take a minute!

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    start_time: left edge of timeblock, string (hh:mm AM/PM)
+    end_time: right edge of timeblock, string (hh:mm AM/PM)
+    swipes: number of swipes, int
+    brb_only: indicates if this location is a brb only eatery (used for calcs), boolean
+    dining_hall: indicates if a location is a dining hall (used for calcs), boolean
+    trillium: indicates if a location is Trillium (used for calcs), boolean
+  """
+  global breaks
+  global weekdays
+  MAIN_CSV = '{}{}'.format(EATERY_DATA_PATH, file_name)
```
young-k: this isn't a constant, so it shouldn't be caps

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+      # check if data csv file already exists
+      if isfile(MAIN_CSV):
+        try:
+          df = pd.read_csv(MAIN_CSV)
+          marker_row = df.head(1)
+          print(marker_row.iloc[0]['date'])
+          df.drop(marker_row.index, inplace=True)  # remove previous time marker
+          update_info['end'] = marker_row
+          update_info['old_data'] = df
+          print('marker and old data found')
```
young-k: remove `prints` if they're unnecessary?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        except Exception as e:
+          print('failed in marker row creation')
+
+      # read most recent data first
+      for line in list(reversed(swipe_data.readlines())):
+        if line_counter == limit:  # limit used for testing purposes
+          break
+        # skip over empty log lines (odd lines going forward, even lines goind reversed)
```
young-k: typo on `goind`?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+          df.drop(marker_row.index, inplace=True)  # remove previous time marker
+          update_info['end'] = marker_row
+          update_info['old_data'] = df
+          print('marker and old data found')
+        except Exception as e:
+          print('failed in marker row creation')
+
+      # read most recent data first
+      for line in list(reversed(swipe_data.readlines())):
+        if line_counter == limit:  # limit used for testing purposes
```
young-k: should this be removed then, if it's out of testing?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+          print(marker_row.iloc[0]['date'])
+          df.drop(marker_row.index, inplace=True)  # remove previous time marker
+          update_info['end'] = marker_row
+          update_info['old_data'] = df
+          print('marker and old data found')
+        except Exception as e:
+          print('failed in marker row creation')
+
+      # read most recent data first
+      for line in list(reversed(swipe_data.readlines())):
```
young-k: hm it's not good to call `readlines()`, since you don't want to overload memory, it's much better to iterate line by line!

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        if line_counter == limit:  # limit used for testing purposes
+          break
+        # skip over empty log lines (odd lines going forward, even lines goind reversed)
+        if i % 2 == 0 or line == '\n':
+          i += 1
+          continue
+        obj = json.loads(line)
+        timestamp_str = obj['TIMESTAMP']
+        if timestamp_str == 'Invalid date' or obj['UNITS'] == []:
```
young-k: can you add a comment explaining `obj['UNITS']`, also you can check an empty list by doing `or not obj['UNITS']`

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+          i += 1
+          line_counter += 1
```
young-k: are both `line_counter` and `i` necessary? and can `i` be named to something more verbose?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+          continue
+        if 'new_marker' not in update_info:
+          # mark the timeblock to stop at on next update
```
young-k: `timeblock`? should this be `timestamp`?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        if 'end' in update_info and update_info['end'].iloc[0]['date'] == timestamp_str:
+          print('hit time marker from last update on {}'.format(timestamp_str))
+          break
+        try:
+          timestamp = datetime.strptime(timestamp_str, '%Y-%m-%d %I:%M:00 %p')
+        except:
+          try:
+            timestamp = datetime.strptime(timestamp_str, '%Y-%m-%d %I:%M:01 %p')
+          except:
+            continue
```
young-k: this is not good, we should never nest `try-except` clauses

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        date = timestamp.date()
+        # print(timestamp_str)
```
young-k: delete rather than comment

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        session_type = sort_session_type(date, breaks)
+        if session_type != 'regular':
+          print(session_type)
```
young-k: is this necessary or can it be deleted?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+      print('done parsing data.log')
+      if 'old_data' in update_info:
+        data_list.append(update_info['old_data'])
+      df = pd.concat(data_list)
+      df.to_csv(MAIN_CSV, header=True, index=False)
+      data_file = sort_by_timeblock(MAIN_CSV)
+      # add new marker to front of table to be used next time for time tracking
+      df = pd.concat([update_info['new_marker'], df])
+      df.to_csv(MAIN_CSV, header=True, index=False)
+      return data_file
```
young-k, line 152: there needs to be more whitespace in the above function so it's easier to follow, add newlines between different sections of code!

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    df = pd.read_csv(input_file_path)
+    # sum together the swipes of rows with same timeblock/eatery makeup
+    df = df.groupby(ISOLATE_SWIPES).sum().reset_index()
+    # count how many days of the same timeblock/eatery combo we have --> used for counter column
+    df = df.groupby(ISOLATE_DATE).count().reset_index()
+    df = df.rename(index=str, columns={'date': 'counter'})
+    # sum together swipes and counters for rows with the same timeblock and location
+    df = df.groupby(ISOLATE_COUNTER_SWIPES).sum().reset_index()
+    df = calculate_wait_times(df)
+    df = df.sort_values(by=['location', 'weekday'])
```
young-k: writing to the same variable over and over again is no bueno, can these variable names be more verbose

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    # sum together the swipes of rows with same day/eatery makeup
+    df = df.groupby(['date', 'session_type', 'weekday', 'location']).sum().reset_index()
+    # count the number of times a location/date pair occur
+    df = df.groupby(['weekday', 'location', 'swipes', 'session_type']).count().reset_index()
+    df = df.rename(index=str, columns={'date': 'counter'})
+    # sum swipes and counters for rows with the same weekday and location
+    df = df.groupby(['weekday', 'location', 'session_type']).sum().reset_index()
+    df['average'] = np.around(np.divide(df['swipes'], df['counter']), decimals=2)
+    df = df.sort_values(by=['location', 'weekday'])
+    df.to_csv('{}{}'.format(EATERY_DATA_PATH, output_file), header=True, index=False)
```
young-k: same comment

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    print('Data update failed:', e)
+
+
+def export_data(file_path):
+  """
+  Transforms our tabular data into custom objects to be placed in Eatery objects
+  """
+  global breaks
+  global weekdays
+  try:
```
young-k, line 248: this function needs more newlines!

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    df = pd.read_csv(file_path)
+    data = {}
+    today = date.today()
+    session_type = sort_session_type(today, breaks)
+    print(weekdays[today.weekday()])
```
young-k: delete if unnecessary! also a great place for a newline is before a `for` loop

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    for location in df['location'].unique():
+      true_location = LOCATION_NAMES[location]['name']
```
young-k: `true_location`? what makes it a "true" location?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+def export_data(file_path):
+  """
+  Transforms our tabular data into custom objects to be placed in Eatery objects
+  """
+  global breaks
+  global weekdays
+  try:
+    df = pd.read_csv(file_path)
```
young-k, line 249: make `df` more verbose

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+    data = {}
+    today = date.today()
+    session_type = sort_session_type(today, breaks)
+    print(weekdays[today.weekday()])
+    for location in df['location'].unique():
+      true_location = LOCATION_NAMES[location]['name']
+      # look at information that pertains to today's criteria
+      new_df = df.loc[(df['location'] == location) & (df['weekday'] == weekdays[today.weekday()])]
```
young-k: maybe `today_df`?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+        new_df = new_df.loc[(new_df['session_type'] == 'regular')]
+        # new_df = aggregate_breaks(new_df)
+      max_swipes = new_df['average'].max()
+      json_data = json.loads(new_df.to_json(orient='table'))
+      for row in json_data['data']:
+        new_timeblock = SwipeDataType(
+            end_time=row['end_time'],
+            session_type=session_type,
+            start_time=row['start_time'],
+            swipe_density=round(row['average'] / max_swipes, 3),
```
young-k: why 3? should this be in a constant?

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+
+def wait_time_multiply(df, type, multiplier):
+  return np.multiply(df['average'], multiplier, where=df[type])
+
+def sort_session_type(date, breaks):
+  """
+  sorts the given date to the proper session_type
+  """
+  for label, dates in breaks.items():
+    if dates[0] <= date <= dates[1]:
```
young-k, line 340: i did not know you could do this, pretty nifty

src/swipes.py
```diff
@@ -0,0 +1,304 @@
...
+      if new_df['session_type'].str.contains(session_type, regex=False).any():
+        new_df = new_df.loc[(new_df['session_type'] == session_type)]
+      else:  # not all breaks will be in data to start, default to regular for time being
+        print('defaulting to regular session_type')
+        new_df = new_df.loc[(new_df['session_type'] == 'regular')]
+        # new_df = aggregate_breaks(new_df)
+      max_swipes = new_df['average'].max()
+      json_data = json.loads(new_df.to_json(orient='table'))
+      for row in json_data['data']:
+        new_timeblock = SwipeDataType(
```
young-k: maybe change from `timeblock`, or explain what a `timeblock` is?

young-k (commented):
More comments, feel free to DM me if you have any questions!

src/constants.py
ilikesymmetry: not sure what you mean by 'plural'? I called it ISOLATE_DATE because it contains all of the columns except for dates @young-k

src/constants.py
ilikesymmetry: this is used for determining whether or not a swipe event occurs during regular school conditions, or over a break.  This discrepancy is used to keep the wait time estimates as accurate as possible.  You are correct in that this would require updates yearly and I want to find a way to have it auto update by webscraping or something @young-k

src/swipes.py
ilikesymmetry: I agree that redoing the same variable names is no bueno, This is just how I saw others doing it in examples.  I'll read for better way

src/swipes.py
ilikesymmetry: like the actual name that makes sense (ex: 'Duffield' --> 'Mattin's Cafe')

src/swipes.py
ilikesymmetry: Idk just chose 3 because digits past this will have little significance when rendering the bars

ilikesymmetry:
just made my first round of changes, left unresolved conversations as things still left to change :)

src/constants.py
meganle: I could do the web scraping or point you in the right direction! I wrote a scraper for Uplift so we can talk about this

ilikesymmetry:
@young-k @meganle @orkosinha @sahamanish99 I made heller changes.  If y'all could give it another review pls :)

src/swipes.py
```diff
@@ -0,0 +1,337 @@
...
+          df.drop(marker_row.index, inplace=True)  # remove previous time marker
+          update_info['end'] = marker_row
+          update_info['old_data'] = df
+        except Exception as e:
+          print('failed in marker row creation')
+
+      # read most recent data first
+      for line in swipe_data:
+        # skip over empty log lines
+        if line == '':
```
meganle: You can just do `if not line:`

src/swipes.py
```diff
@@ -0,0 +1,337 @@
...
+
+  try:
+    df = pd.read_csv(file_path)
+    data = {}
+    today = date.today()
+    session_type = sort_session_type(today, breaks)
+    for location in df['location'].unique():
+      eatery_name = LOCATION_NAMES[location]['name']
+      # look at information that pertains to today's criteria
+      today_df = df.loc[(df['location'] == location) & (df['weekday'] == weekdays[today.weekday()])]
```
meganle, line 256: `and` instead of `&` ?

src/swipes.py
```diff
@@ -0,0 +1,337 @@
...
+        print('{} has no swipe data for {}'.format(eatery_name, weekdays[today.weekday()]))
+        continue
+
+      # df contains current session
+      if today_df['session_type'].str.contains(session_type, regex=False).any():
+        today_df = today_df.loc[(today_df['session_type'] == session_type)]
+      else:  # not all breaks will be in data to start, default to regular for time being
+        print('defaulting to regular session_type')
+        today_df = today_df.loc[(today_df['session_type'] == 'regular')]
+        # today_df = aggregate_breaks(today_df)
```
meganle: Can this be deleted?

meganle (approved):
LGTM! Just some last minor fixes

src/swipes.py
ilikesymmetry: 👍

src/swipes.py
ilikesymmetry: actually we need the `&` for pandas df.  just how it works

src/swipes.py
ilikesymmetry: sure

src/constants.py
```diff
@@ -3,39 +3,83 @@
...
+  'North Star Marketplace': {'name': 'North Star Dining Room', 'type': DINING_HALL},
+  'Okenshields': {'name': 'Okenshields', 'type': DINING_HALL},
+  'Olin Libe Cafe': {'name': 'Amit Bhatia Libe Café', 'type': BRB_ONLY},
+  'RPME': {'name': 'Robert Purcell Marketplace Eatery', 'type': DINING_HALL},
+  'Risley': {'name': 'Risley Dining Room', 'type': DINING_HALL},
+  'Rose House': {'name': 'Rose House Dining Room', 'type': DINING_HALL},
+  'Rustys': {'name': "Rusty's", 'type': BRB_ONLY},
+  'Sage': {'name': 'Atrium Café', 'type': BRB_ONLY},
+  'Statler Macs': {'name': "Mac's Café", 'type': BRB_ONLY},
+  'Statler Terrace': {'name': 'The Terrace', 'type': BRB_ONLY},
```
young-k: why is indentation 2 here but 4 elsewhere in this file?

src/schema.py
young-k: bump on this @cswenberg, you can resolve if it's not possible but i don't think we should have two rounds of parsing –– if anything, just put it in an `if else` block and it would probably be cleaner

src/constants.py
```diff
@@ -3,39 +3,83 @@
...
+    'session_type',
+    'weekday',
+    'location',
+    'start_time',
+    'end_time',
+    DINING_HALL,
+    BRB_ONLY,
+    TRILLIUM
+]
+ISOLATE_DATE = [
```
young-k: i know you explained this, but the `ISOLATE` variables aren't intuitive to someone reading the code for the first time, the `ISOLATE` variables either need comments or better names. 

Also, why not define one list that has the union of these elements, and then build the three `ISOLATE` lists by copying the master list, and removing the unnecessary ones?

src/swipes.py
```diff
@@ -0,0 +1,336 @@
...
+
+        if 'end' in update_info and update_info['end'].iloc[0]['date'] == timestamp_str:
+          print('hit time marker from last update on {}'.format(timestamp_str))
+          break
+
+        try:
+          try:
+            timestamp = datetime.strptime(timestamp_str, '%Y-%m-%d %I:%M:00 %p')
+          except:
+            timestamp = datetime.strptime(timestamp_str, '%Y-%m-%d %I:%M:01 %p')
```
young-k: same comment on nested `try` `except` clauses here

src/swipes.py
```diff
@@ -0,0 +1,336 @@
...
+    df['wait_time_low'] = wait_time_multiply(df, type, multiplier)
+    df['wait_time_high'] = wait_time_multiply(df, type, multiplier)
+  df['wait_time_low'] = np.floor(df['wait_time_low'])
+  df['wait_time_high'] = np.ceil(df['wait_time_high'])
+  df['wait_time_high'] = np.add(df['wait_time_high'], 2)  # expands bounds
+  print('wait time calculations done')
+  return df
+
+def wait_time_multiply(df, type, multiplier):
+  """Returns a new columns"""
```
young-k: `a new columns`?

young-k (approved):
LGTM, minor comments

src/swipes.py
ilikesymmetry: ooof my english broke for a sec oops.

src/constants.py
ilikesymmetry: ooh I dig that idea

PR #94, opened 2019-04-19
Get hours not on Yelp from an external source

orkosinha:
Added a json file to DiningStack which is now being read to populate hours instead of Yelp (because Yelp doesn't have them or is incorrect).


ilikesymmetry (commented):
lgtm :)

ilikesymmetry (approved):

meganle (approved):
nice!!

PR #98, opened 2019-04-22
Change wait time multiplier mechanism

ilikesymmetry:
Rids us of needing a column for each multiplier we have in the table

### DIFF
--- a/src/eatery/static_eatery.py
+++ b/src/eatery/static_eatery.py
@@ -115,4 +115,4 @@ def string_to_date_range(dates):
   start_str, end_str = dates.split('-')
   start_date = datetime.strptime(start_str, '%m/%d/%y').date()
   end_date = datetime.strptime(end_str, '%m/%d/%y').date()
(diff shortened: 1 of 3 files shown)


src/eatery/static_eatery.py
```diff
@@ -115,4 +115,4 @@ def string_to_date_range(dates):
   start_str, end_str = dates.split('-')
   start_date = datetime.strptime(start_str, '%m/%d/%y').date()
   end_date = datetime.strptime(end_str, '%m/%d/%y').date()
-  return [start_date, end_date]
+  return (start_date, end_date)
```
kq-li, line 118: i'd personally remove the parens, @young-k?

kq-li (approved):
tight

meganle (approved):
Good catch 🎣

src/eatery/static_eatery.py
young-k: i don't really mind either way, up to @meganle

PR #99, opened 2019-04-24
Increase gunicorn timeout

meganle:
Parsing the swipe data takes a while so when the server starts the workers would timeout before it could finish so the workers would reboot and retry. Increasing the timeout allows sufficient time for the workers to finish parsing the swipe data (tested locally and is currently running in prod)


ilikesymmetry (approved):
yeeet lets gooo

orkosinha (approved):
yeet

PR #100, opened 2019-04-25
Add expanded menu

orkosinha:
Added an expandedMenu field to the CampusEateryType.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -8,6 +8,7 @@ BRB_ONLY = 'brb_only'
 CORNELL_DINING_URL = 'https://now.dining.cornell.edu/api/1.0/dining/eateries.json'
 CORNELL_INSTITUTION_ID = '73116ae4-22ad-4c71-8ffd-11ba015407b1'
 DINING_HALL = 'dining_hall'
+DINING_STACK_URL = 'https://raw.githubusercontent.com/cuappdev/DiningStack/master/DiningStack/'
 EATERY_DATA_PATH = './eatery-data/'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
 GIT_CONTENT_URL = 'https://raw.githubusercontent.com/cuappdev'
@@ -67,10 +68,10 @@ SCHOOL_BREAKS = {
   'finals_spring': '5/8/19-5/18/19',
   'summer': '5/19/19-8/28/19'
 }
-STATIC_EATERIES_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalEateries.json'
-STATIC_CTOWN_HOURS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalHours.json'
-STATIC_MENUS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/hardcodedMenus.json'
-STATIC_EXPANDED_ITEMS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/expandedItems.json'
+STATIC_EATERIES_URL = DINING_STACK_URL + 'externalEateries.json'
+STATIC_CTOWN_HOURS_URL = DINING_STACK_URL + 'externalHours.json'
+STATIC_MENUS_URL = DINING_STACK_URL + 'hardcodedMenus.json'
+STATIC_EXPANDED_ITEMS_URL = DINING_STACK_URL + 'expandedItems.json'
 SWIPE_DENSITY_ROUND = 3
 SWIPE_PLANS = ['Bear Basic', 'Bear Choice', 'Bear Traditional']
 TABLE_COLUMNS = [
--- a/src/constants.py
+++ b/src/constants.py
@@ -70,6 +70,7 @@ SCHOOL_BREAKS = {
 STATIC_EATERIES_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalEateries.json'
 STATIC_CTOWN_HOURS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalHours.json'
 STATIC_MENUS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/hardcodedMenus.json'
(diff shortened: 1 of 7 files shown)


src/constants.py
```diff
@@ -70,6 +70,7 @@
 STATIC_EATERIES_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalEateries.json'
 STATIC_CTOWN_HOURS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/externalHours.json'
 STATIC_MENUS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/hardcodedMenus.json'
+STATIC_EXPANDED_ITEMS_URL = GIT_CONTENT_URL + '/DiningStack/master/DiningStack/expandedItems.json'
```
ilikesymmetry: we should make a dining stack url, looking at this one and the few above, we are repeating ourselves a lot

src/eatery/common_eatery.py
```diff
@@ -101,6 +108,82 @@ def parse_events(event_list, event_date):
...
+
+  Args:
+    eatery (dict): A valid json segment from Cornell Dining that contains eatery information
+  """
+  eatery_slug = eatery['slug']
+  menus = requests.get(STATIC_EXPANDED_ITEMS_URL).json()
+  categories = []
+
+  for menu in menus['eateries']:
+    if menu['slug'] == eatery_slug:
```
ilikesymmetry: since this is the only use of `eatery_slug`, we can just put `eatery['slug']` in place instead of using a one-time variable

src/eatery/common_eatery.py
```diff
@@ -101,6 +108,82 @@ def parse_events(event_list, event_date):
...
+  Returns a list of DescriptiveFoodItemTypes for an eatery's station.
+
+  Args:
+    eatery_items (dict): A valid json segment from the hard-coded menu items
+  """
+  items = []
+  for item in eatery_items:
+    food_item = DescriptiveFoodItemType(
+      item=item['item'],
+      healthy=False,
```
ilikesymmetry, line 163: this should have some actual logic in the future, but for v1 this is fine.  Maybe drop a comment that this should be changed in the future.

src/eatery/common_eatery.py
```diff
@@ -101,6 +108,82 @@ def parse_events(event_list, event_date):
...
+      options=parse_expanded_item_options(item['choices']),
```
ilikesymmetry: should keep consistent naming across the json file and call this `choices`

src/types/__init__.py
```diff
@@ -6,6 +6,12 @@
     CoordinatesType,
     RatingEnum
 )
+from .expanded_menu_type import (
+    FoodCategoryType,
```
ilikesymmetry: unfortunately, F does not come before D in the alphabet :(

src/types/expanded_menu_type.py
```diff
@@ -0,0 +1,24 @@
...
+
+class DescriptiveFoodItemType(FoodItemType):
+  options = List(DescriptiveFoodItemOptionType, required=False)
+  price = String(required=True)
+
+  def equals(self, food_item):
+    if len(self.options) == len(food_item.options):
+      return False
+    return (super(FoodItemType, self).equals(self, food_item) and self.price == food_item.price
+            and all([self.options[i].equals(food_item.options[i]) for i in range(self.options)]))
```
ilikesymmetry, line 18: need a comment to make this easier for someone to get what's going on

src/types/expanded_menu_type.py
```diff
@@ -0,0 +1,24 @@
...
+  label = String(required=True)
+  options = List(String, required=True)
+
+class DescriptiveFoodItemType(FoodItemType):
+  options = List(DescriptiveFoodItemOptionType, required=False)
+  price = String(required=True)
+
+  def equals(self, food_item):
+    if len(self.options) == len(food_item.options):
+      return False
```
ilikesymmetry, line 15: why would we return false if they the number of options are the same?

src/types/expanded_menu_type.py
```diff
@@ -0,0 +1,24 @@
+from graphene import List, ObjectType, String
+
+from src.types.food_station_type import FoodItemType, FoodStationType
+
+class DescriptiveFoodItemOptionType(ObjectType):
+  label = String(required=True)
+  options = List(String, required=True)
+
+class DescriptiveFoodItemType(FoodItemType):
+  options = List(DescriptiveFoodItemOptionType, required=False)
```
ilikesymmetry: same note as in parsing, should keep naming conventions consistent with json file.

ilikesymmetry (changes requested):
all in all, nice addition.  just a few small things to change 👍

src/eatery/common_eatery.py
```diff
@@ -1,13 +1,20 @@
+import requests
```
ilikesymmetry: put under the datetime import

src/eatery/common_eatery.py
```diff
@@ -101,6 +108,82 @@ def parse_events(event_list, event_date):
...
+    item_options (dict): A valid json segment from the hard-coded menu items
+  """
+  option_categories = []
+  for option in item_options:
+    item_option = DescriptiveFoodItemOptionType(
+      label=option['label'],
+      options=option['options']
+    )
+    option_categories.append(item_option)
+  return option_categories
```
young-k: Perhaps a more nifty way to write all of these functions, and more clear as well:

```python
option_to_category = lambda option: DescriptiveFoodItemOptionType(
  label=option['label'],
  options=option['options']
)
return map(option_category, item_options)
```

@cswenberg @meganle thoughts?

ilikesymmetry (commented):
lgtm :)

ilikesymmetry (approved):

src/eatery/common_eatery.py
ilikesymmetry: I'm not familiar with python's `lambda` features but if it works, it works. @young-k

orkosinha: I followed the conventions of a lot of the other parsing functions in the code, we could potentially have all of them use anonymous functions?

young-k (approved):
#shipit, let's add a separate PRs to use lambdas because I think it's more concise and clean to read @orkosinha

PR #102, opened 2019-05-05
Fix GET naming scheme and table manipulation script

ilikesymmetry:
Modularizes the GET naming fixes and catches more locations.  Also wrote a script to manually change the formatting of our start and end times to help out frontend.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -11,13 +11,21 @@ DINING_HALL = 'dining_hall'
 DINING_STACK_URL = 'https://raw.githubusercontent.com/cuappdev/DiningStack/master/DiningStack/'
 EATERY_DATA_PATH = './eatery-data/'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
+GET_LOCATIONS = {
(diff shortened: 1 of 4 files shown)


src/constants.py
```diff
@@ -11,13 +11,21 @@
 DINING_STACK_URL = 'https://raw.githubusercontent.com/cuappdev/DiningStack/master/DiningStack/'
 EATERY_DATA_PATH = './eatery-data/'
 GET_URL = 'https://services.get.cbord.com/GETServices/services/json'
+GET_LOCATIONS = {
+    'Attrium Cafe': 'Atrium Café',
```
kq-li, line 15: is this supposed to have 2 t's?

kq-li (approved):
lgtm!

src/constants.py
ilikesymmetry: no, that's why we change it

kq-li: ahh thought it was a typo, all good

orkosinha (approved):
lgtm!

PR #104, opened 2019-05-13
Remove Sweetspot from collegetown

ilikesymmetry:
Also includes a recommended package update for `urllib3`.  Additional error handling was including to prevent the complete absence of ctown information if not every query works properly.

### DIFF
--- a/src/collegetown.py
+++ b/src/collegetown.py
@@ -4,6 +4,7 @@ from time import sleep
 from yelpapi import YelpAPI
 
 from src.constants import (
+    SWEETSPOT_ID,
     YELP_LATITUDE,
     YELP_LONGITUDE,
     YELP_QUERY_DELAY,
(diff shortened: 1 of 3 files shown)


src/collegetown.py
```diff
@@ -21,14 +22,23 @@ def collegetown_search():
...
-    eateries = []
-    for place in search['businesses']:
-      eatery_id = place['id']
-      id_query = yelp_api.business_query(id=eatery_id)
-      eateries.append(id_query)
-      sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
-    print('Done with Collegetown Search')
-    return eateries
   except Exception as e:
+    print('failed at yelp radial search')
```
young-k, line 26: prefer that this is capitalized, `Failed at Yelp radial search`

src/collegetown.py
```diff
@@ -21,14 +22,23 @@ def collegetown_search():
...
     print(e)
     return []
+
+  eateries = []
+  for place in search.get('businesses', []):
```
young-k, line 31: why not:

```python
if 'businesses' not in search:
  return []

eateries = []
for place in search['businesses']:
  ...
```

also, is `businesses` actually optional here?

src/collegetown.py
```diff
@@ -21,14 +22,23 @@ def collegetown_search():
...
+    eatery_id = place['id']
+    if eatery_id == SWEETSPOT_ID:
+      continue
+    try:
+      id_query = yelp_api.business_query(id=eatery_id)
+    except Exception as e:
+      print('failed at business query with id {}'.format(eatery_id))
```
young-k, line 38: `Failed at business query...`

src/collegetown.py
```diff
@@ -21,14 +22,23 @@ def collegetown_search():
...
+      print(e)
+      continue
+    eateries.append(id_query)
+    sleep(YELP_QUERY_DELAY)  # delay between query calls to not trigger yelp query limits
```
young-k, line 42: we should not `sleep` in code, because that shuts down the entire process and makes it unresponsive, this should be in a thread instead!!!

src/constants.py
```diff
@@ -78,6 +78,7 @@
 STATIC_CTOWN_HOURS_URL = DINING_STACK_URL + 'externalHours.json'
 STATIC_MENUS_URL = DINING_STACK_URL + 'hardcodedMenus.json'
 STATIC_EXPANDED_ITEMS_URL = DINING_STACK_URL + 'expandedItems.json'
+SWEETSPOT_ID = '8FpVJek1uqdaowPLZptfWg'
```
young-k, line 81: add a comment here as to why this is necessary, just so it's clear for the future

young-k (changes requested):
If there are any instances of `sleep` in the code, they need to be removed!

PR #110, opened 2019-08-29
Fix swipe data parsing error

ilikesymmetry:
Encountered a bug where data.log (stores the queries from cornell dining
requests every 5 minutes) contains two logs of {"msg": "Unauthorized"}
which created a key error when we tried to access 'TIMESTAMP' from the
log. This exception was not caught and prevented further updating of
our csv files. This bug may also be responsible for inconsistent ctown
data. We fix the issue by checking the 'TIMESTAMP' is a valid key in the
log.

### DIFF
--- a/src/swipes.py
+++ b/src/swipes.py
@@ -69,7 +69,10 @@ def parse_to_csv(file_name='data.csv'):
           continue
 
         obj = json.loads(line)
-        timestamp_str = obj['TIMESTAMP']
+        timestamp_str = obj.get('TIMESTAMP', None)
+        if not timestamp_str:
+          # we have received {"msg":"Unauthorized"} as a log twice instead of real data
+          continue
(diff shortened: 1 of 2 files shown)


orkosinha (approved):
lgtm

meganle (approved):

src/swipes.py
```diff
@@ -69,7 +69,11 @@ def parse_to_csv(file_name='data.csv'):
           continue
 
         obj = json.loads(line)
-        timestamp_str = obj['TIMESTAMP']
+        timestamp_str = obj.get('TIMESTAMP', None)
+        if not timestamp_str:
+          # we have received {"msg":"Unauthorized"} as a log twice instead of real data
+          print(obj)
```
young-k: Should this be removed?

ilikesymmetry: yeahhhh

Issue #120, opened 2019-09-04
Pad BRB History amount item with two zeros


whoiswillma:
Currently the iOS frontend receives the history amount as a string. Although it is a worthwhile discussion to have as to whether to change the history amount to a numeric value, for now the backend should pad the amount string with two 0's. 

![IMG_1786](https://user-images.githubusercontent.com/1882991/64293934-a8970380-cf3b-11e9-945a-d91d887321c0.jpg)

orkosinha:
Fixed in #123

PR #123, opened 2019-09-06
Add padding to BRB values

orkosinha:
Address Issue #120.

### DIFF
--- a/src/schema.py
+++ b/src/schema.py
@@ -159,7 +159,7 @@ class Query(ObjectType):
         name = location
 
       new_transaction = {
-          'amount': txn['amount'],
+          'amount': "{:.2f}".format(int(txn['amount'])),
           'name': name,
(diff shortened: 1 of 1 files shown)


src/schema.py
```diff
@@ -159,7 +159,7 @@ def resolve_account_info(self, info, session_id=None):
         name = location
 
       new_transaction = {
-          'amount': txn['amount'],
+          'amount': "{:.2f}".format(int(txn['amount'])),
```
young-k: Doesn't casting to `int` lose the precision of the float?

young-k (changes requested):

src/schema.py
orkosinha: Yep, you're right

young-k (approved):

meganle (approved):

PR #124, opened 2019-09-08
Check if Transactions are Positive

orkosinha:
Add some logic to check if a transaction is adding BRBs.

### DIFF
--- a/src/schema.py
+++ b/src/schema.py
@@ -158,9 +158,13 @@ class Query(ObjectType):
       else:
         name = location
 
+      # Transaction Type 3 denotes payments that increase balance
+      positive = txn['transactionType'] == 3
+
       new_transaction = {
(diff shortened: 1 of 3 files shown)


johnathananderson (approved):

src/schema.py
```diff
@@ -158,9 +158,13 @@ def resolve_account_info(self, info, session_id=None):
       else:
         name = location
 
+      # Transaction Type 3 denotes payments that increase balance
+      positive = txn['transactionType'] == 3
```
meganle: let's make a constant for this. like `POSITIVE_TRANSACTION_TYPE = 3` or something similar and add it to the constants.py!

meganle (changes requested):

meganle (approved):
sweet

PR #125, opened 2019-09-11
Error catching for get swipes

orkosinha:
Solves the issue found in server logs. 

```
An error occurred while resolving field Query.accountInfo
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/graphql/execution/executor.py", line 447, in resolve_or_error
    return executor.execute(resolve_fn, source, info, **args)
  File "/usr/local/lib/python3.6/site-packages/graphql/execution/executors/sync.py", line 16, in execute
    return fn(*args, **kwargs)
  File "/usr/src/app/src/schema.py", line 118, in resolve_account_info
    if (int(account_info['swipes']) > 50):
ValueError: invalid literal for int() with base 10: 'None'
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/graphql/execution/executor.py", line 447, in resolve_or_error
    return executor.execute(resolve_fn, source, info, **args)
  File "/usr/local/lib/python3.6/site-packages/graphql/execution/executors/sync.py", line 16, in execute
    return fn(*args, **kwargs)
  File "/usr/src/app/src/schema.py", line 118, in resolve_account_info
    if (int(account_info['swipes']) > 50):
graphql.error.located_error.GraphQLLocatedError: invalid literal for int() with base 10: 'None'
```

I'm not sure which meal plan returns `"None"` swipes and when this happens, but this should catch it.

### DIFF
--- a/src/schema.py
+++ b/src/schema.py
@@ -116,8 +116,11 @@ class Query(ObjectType):
         account_info['swipes'] = str(acct['balance'])
 
     # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
-    if (int(account_info['swipes']) > 50):
-      account_info['swipes'] = 'Unlimited'
+    try:
+      if (int(account_info['swipes']) > 50):
+        account_info['swipes'] = 'Unlimited'
+    except:
+      account_info['swipes'] = 'None'
 
     # Query 3: Get list of transactions
     transactions = requests.post(
--- a/src/schema.py
+++ b/src/schema.py
@@ -115,12 +115,11 @@ class Query(ObjectType):
       elif any(meal_swipe_name in acct['accountDisplayName'] for meal_swipe_name in SWIPE_PLANS):
         account_info['swipes'] = str(acct['balance'])
 
-    # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
-    try:
+    # Check if the balance provided by Cornell Dining is a regular digit
+    if (account_info['swipes'].isdigit()):
+      # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
       if (int(account_info['swipes']) > 50):
         account_info['swipes'] = 'Unlimited'
-    except:
-      account_info['swipes'] = 'None'
 
     # Query 3: Get list of transactions
     transactions = requests.post(


meganle (approved):
LGTM

johnathananderson (approved):

src/schema.py
```diff
@@ -116,8 +116,11 @@ def resolve_account_info(self, info, session_id=None):
         account_info['swipes'] = str(acct['balance'])
 
     # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
-    if (int(account_info['swipes']) > 50):
-      account_info['swipes'] = 'Unlimited'
+    try:
+      if (int(account_info['swipes']) > 50):
+        account_info['swipes'] = 'Unlimited'
+    except:
```
yuna878: Specify the exception type? Also, can you put a comment on when `account_info['swipes'] = None` occurs?

manishsaha (approved):
swag

src/schema.py
orkosinha: I changed it so instead of error checking it checks if the number is a digit. I think this is a better, more efficient solution.

src/schema.py
```diff
@@ -115,9 +115,11 @@ def resolve_account_info(self, info, session_id=None):
...
         account_info['swipes'] = str(acct['balance'])
 
-    # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
-    if (int(account_info['swipes']) > 50):
-      account_info['swipes'] = 'Unlimited'
+    # Check if the balance provided by Cornell Dining is a regular digit
+    if (account_info['swipes'].isdigit()):
+      # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
+      if (int(account_info['swipes']) > 50):
+        account_info['swipes'] = 'Unlimited'
```
yuna878, line 122: Do we need a `else` clause to reset the field to a string value when `swipes` is not a digit?

orkosinha: Nah, the value would just be whatever Cornell Dining gives us if the value isn't a digit. So, in the `"None"` case it would still return `"None"` but without crashing.

yuna878 (approved):
Looks good!

src/schema.py
```diff
@@ -115,9 +115,11 @@ def resolve_account_info(self, info, session_id=None):
       elif any(meal_swipe_name in acct['accountDisplayName'] for meal_swipe_name in SWIPE_PLANS):
         account_info['swipes'] = str(acct['balance'])
 
-    # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
-    if (int(account_info['swipes']) > 50):
-      account_info['swipes'] = 'Unlimited'
+    # Check if the balance provided by Cornell Dining is a regular digit
+    if (account_info['swipes'].isdigit()):
+      # Check if the meal plan has more than 50 swipes, this is larger than the largest plan.
+      if (int(account_info['swipes']) > 50):
```
young-k, line 121: i know we closed this, but it's non-idiomatic to have parentheses here and in the above `if` statement!

PR #129, opened 2019-09-18
Preload data before starting gunicorn workers

meganle:
After doing some research, I discovered you can do stuff before gunicorn starts its workers!
- Before starting the workers, you can make gunicorn run a method called `on_starting`
- I placed the `start_update()` method call in here which handles swipe data, campus eateries, and collegetown eateries
- That way this data update process doesn't run separately by each worker during startup but only once in the beginning
- Previously a lot of gunicorn workers would timeout and do this process over again making startup time infinite
- I also moved the gunicorn configurations into the `gunicorn.py` file

You can see here that the data fetching occurs before the workers start
Screen Shot 2019-09-18 at 12 23 47 AM

Screen Shot 2019-09-18 at 12 24 40 AM

This is currently deployed on prod since the previous deployed image failed to startup when enabling swipe data. This is a temporary solution until we can get the database up.

### DIFF
--- a/app.py
+++ b/app.py
@@ -7,7 +7,6 @@ from src.schema import Query
 
 app = Flask(__name__)
 schema = Schema(query=Query)
-start_update()
 
 app.add_url_rule(
     '/',
--- a/gunicorn.py
+++ b/gunicorn.py
@@ -1,0 +1,10 @@
+from src.data import start_update
+
+def on_starting(server):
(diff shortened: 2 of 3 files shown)


orkosinha (approved):
Looks super good to me

young-k (approved):
nice stuff

PR #130, opened 2019-09-18
Make swipe dates ISO 8601

orkosinha:
Change date format from `HH:MM` to `yyyy-mm-dd:HH:MM:PM`, as the current Android app relies on this format.

### DIFF
--- a/src/swipes.py
+++ b/src/swipes.py
@@ -116,15 +116,15 @@ def parse_to_csv(file_name="data.csv"):
                 # sort time into a time block
                 if timestamp.minute > 30:
                     delta = timedelta(minutes=30)
-                    start_time = timestamp.strftime("%H:30")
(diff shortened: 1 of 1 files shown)


young-k (commented):
Will this break functionality on iOS?

orkosinha:
Either iOS or Android needs to change how they want to handle swipe data dates

orkosinha:
> Will this break functionality on iOS?
iOS doesn’t make any swipe related queries yet, so it should be good.

orkosinha:
> Will this break functionality on iOS?


I found out later that iOS doesn't use swipe data at all right now, so nothing should change @young-k

meganle (approved):
If nothing breaks, then cool

yuna878 (approved):

PR #131, opened 2019-09-22
Add documentation for types

yuna878:


meganle:
Is this good to be reviewed? You didn't request anyone to review

young-k (changes requested):
Is this change necessary? This is just reiterating the params that are already defined –– it seems like if we were to merge this change, we would just be duplicating this code, making it harder to maintain in the future.

Instead, why not use the docstring to describe what the type is defining?

yuna878:
@meganle Still have a few fields to complete, which I couldn't because the site was down so I couldn't access data. I'll finish this PR up as soon as I'm on tract with all the work
@young-k I thought the multiple types were hard to navigate as I was onboarding and multiple members of the pod wanted better documentation so I thought this could be a start because it would help the developer & same documentation could be used on GraphQL for anyone who is trying to access data through client. What do you mean by using docstring to define types?

PR #136, opened 2019-09-25
Skip lines in swipe logs that aren't formatted correctly

meganle:
Swipe data wasn't working due to a parse error and upon looking at the data.log file I found this in the logs so I added some error handling to skip lines like this

Screen Shot 2019-09-25 at 12 47 16 AM

Tested in locally by downloading the swipe data on the server


orkosinha (approved):
LGTM

yuna878 (approved):

PR #137, opened 2019-09-25
Add error checking to Get requests

orkosinha:
Currently, the transactions method doesn't work on Get. Added error checking to have some Get information sent back.


meganle (approved):
Make sure you run `pre-commit install` in your project directory and that your virtualenv has the updated requirements installed so that black and flake8 runs when you make a commit!

yuna878 (approved):
looks good

PR #138, opened 2019-09-26
Change date in export_data

orkosinha:
This would revert #130 as it changed the formatting of the CSV file. A change to the way we wrote the CSV file would require the entire CSV file to be regenerated. This would change the date to ISO 8601, and preserve the original format of the CSV file.


johnathananderson (approved):
Looks good

meganle (approved):
if you tested locally, it looks good to go!

yuna878 (approved):

PR #140, opened 2019-10-02
Change transactions query

orkosinha:
Change the transactions query to reflect the new method on Get. Should get transactions working on the app again.


johnathananderson (approved):

meganle (approved):
let's ship it!

PR #142, opened 2019-10-09
October hours for Manndibles, Macs, Gimme

connorreinhold:
Updated hours for Mac's: https://www.facebook.com/macscafeterrace/

Updated hours for Manndibles: https://www.manndiblecafe.com/hours.html

For Gimme: https://gimmecoffee.com/locations/

### DIFF
--- a/static_sources/externalEateries.json
+++ b/static_sources/externalEateries.json
@@ -118,7 +118,7 @@
                          {
                              "descr": "General",
                              "start": "7:30am",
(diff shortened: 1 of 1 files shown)


young-k (approved):
Is there a way we could set a reminder to do this monthly? I think `slackbot` has some integrations that can do this for us

orkosinha (approved):
Looks lit

PR #145, opened 2019-10-18
Update Oishii hours

connorreinhold:
Closes at 10pm instead of 11pm. Will only merge once we get confirmation it is 10pm every day.


young-k (approved):
They have their schedule posted on their door, I can check it when I go home today and upload the picture!

orkosinha (approved):
lgtm, just double check the hours

PR #146, opened 2019-10-23
Update dates for school breaks

meganle:
## Overview
Swipe data uses the dates for school breaks to take into account times where traffic is irregular.

## Changes Made
Updated the dates by looking at the registrar's website for the academic calendar for this year and next year

I wrote a scraper to do this automatically but not sure if it's worth using. I wrote most of it last semester but ended up having to make changes since the formatting of the web page changed. Web scraping is pretty unreliable since if Cornell changes the format again, the scaper won't work.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -65,14 +65,14 @@ PAY_METHODS = {
 }
 POSITIVE_TRANSACTION_TYPE = 3
 SCHOOL_BREAKS = {
-    "fall": "10/6/18-10/9/18",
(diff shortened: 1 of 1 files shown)


yuna878 (approved):

orkosinha (approved):
lgtm

yuna878:
If this is something that needs update every year, can we add the dates for next few years? I think the calendars are already up. Or can we set up some kind of alert (maybe on slack) for a reminder?

PR #153, opened 2019-11-07
Change db model base

yuna878:
## Overview

Make more fixes to db models.

## Changes Made

**Problem** : Sqlalchemy's `declarative_base` classes were not being initialized correctly because the individual model classes were inheriting python's `object` class, then being declared `__bases__` attribute later on with `type(name, base)` function. Due to the incorrect setup, the model classes didn't have `__init__` functions (note: `declarative_base` class should automatically have `__init__` functions enabled). 

**Solution** : To solve this problem, the db model classes need to inherit `Base` class when they are initialized. One of the reasons of using the `type(name, base)` function was to get rid of redundancy between overlapping fields for `CampusEatery` and `CollegetownEatery` but there is no way to factor out these fields and have the classes inherit the same `Base` class.

## Test Coverage

Script:
```
// Test creating db tables
from src.db_models.CampusEatery import CampusEatery
from src.db_models.CampusEateryHour import CampusEateryHour
from src.db_models.CollegetownEatery import CollegetownEatery
from src.db_models.CollegetownEateryHour import CollegetownEateryHour
from src.db_models.ExpandedMenuChoice import ExpandedMenuChoice
from src.db_models.ExpandedMenuItem import ExpandedMenuItem
from src.db_models.ExpandedMenuStation import ExpandedMenuStation
from src.db_models.MenuCategory import MenuCategory
from src.db_models.MenuItem import MenuItem
from src.db_models.SwipeData import SwipeData
from src.db import Base, db_session, engine

Base.metadata.create_all(bind=engine)

// Test performing db operations
c = CampusEatery(
    about="test",
    campus_area_desc="test",
    eatery_type="test",
    image_url="test",
    latitude=12.112,
    location="test",
    longitude=12.121,
    name="test",
    name_short="test",
    payment_method_brbs=True,
    payment_method_cash=True,
    payment_method_cornell_card=True,
    payment_method_credit=True,
    payment_method_mobile=True,
    payment_method_swipes=True,
    phone="test",
    slug="test",
)
db_session.add(c)
db_session.commit()
```

Result: 
Screen Shot 2019-11-06 at 6 19 35 PM


meganle (approved):
🐟

kevinchan159 (approved):
🐟

Omarrasheed (approved):
🐟

orkosinha (approved):
Looks 🔥

PR #154, opened 2019-11-10
[Populate DB][1/3] Add Cornell Dining eateries to database

orkosinha:
## Overview

Add all of the Cornell Dining eateries to our database. This doesn't add any Swipe Data for Cornell Dining, but it completely adds all other fields. Collegetown eatery tables are also partially implemented, none of the hours are included. The expanded menu tables also not populated.

## Changes Made

I had to rename the `types` folder as it was causing a conflict with SQLAlchemy.
I also rearranged some files to make `db.py` function similarly to `data.py` with a `start_update()` function that populated the database. 

## Test Coverage

I manually tested populating the database using a SQL DB Browser and compared it with the current GraphQL results.

## Next Steps

The next steps are to implement the rest of Collegetown eateries. Additionally, swipe data should also be populated in the database. I'm kind of unsure about expanded menu though, as it is hardcoded data that has never been used by the app before. It's also not looking like this feature will be implemented, so it may be better to remove it altogether. 

## Screenshots 

<details>

  <summary>Campus Eateries</summary>
campus_eatery
campus_eatery_hours
menu_categories
menu_items

  <summary>Collegetown Eateries</summary>
collegetown

  
  

</details>


src/db.py
```diff
@@ -1,20 +1,56 @@
...
+    parse_menu_categories,
+    parse_menu_items,
+)
 
-engine = create_engine("sqlite:///data.sqlite3", convert_unicode=True)
-db_session = scoped_session(sessionmaker(autocommit=False, autoflush=False, bind=engine))
 
+def start_update():
+    try:
+        Base.metadata.drop_all(bind=Engine)
```
meganle, line 20: rather than dropping all of the tables, maybe we could just drop the tables that contain the menu and hours? since the other data should remain the same

yuna878: If we do that, we won't know when cornell dining changes some general information about the eateries - e.g. changing payment methods (idk if this will be a huge issue though)

yuna878 (approved):
👍

src/db.py
orkosinha: I don't think the basic eatery information is refreshed that often, so I agree that it shouldn't be refreshed in our database every day.

meganle (approved):
Nice, looks good now! Quick question, is there any calls to `start_update` where refresh is True?

PR #156, opened 2019-11-20
Finish database population

orkosinha:
## Overview

This completes the database population aspect of migration to SQLAlchemy. All of `ExpandedMenu`, `SwipeData`, `CampusEateries`, `Collegetown` are populated with data previously stored in dictionaries. 

## Changes Made

Similar to #154 with added functionality for `ExpandedMenu` and `SwipeData`.
Also deleted a duplicate entry for `Fork and Gavel Cafe`.

## Test Coverage

Also similar to #154 where I checked on data with DB Browser for SQL Lite.

## Related PRs 
#154


src/db.py
```diff
@@ -27,17 +35,74 @@ def get_campus_eateries(data_json, refresh=False):
...
+                    eatery_categories = (x[0] for x in categories_and_items)
+                    Session.add_all(eatery_categories)
+                    Session.commit()
+
+                    for menu_category, items_json in categories_and_items:
+                        menu_items = parse_menu_items(items_json, menu_category)
+                        Session.add_all(menu_items)
+                        Session.commit()
+
+                campus_eateries.remove(eatery)
```
meganle: why do you remove `eatery` from `campus_eateries`

src/eatery_db/collegetown_eatery.py
```diff
@@ -39,3 +45,43 @@ def parse_collegetown_eateries(collegetown_data):
         collegetown_eateries.append(new_eatery)
 
     return collegetown_eateries
+
+
+def parse_collegetown_hours(data_json, eatery_model):
```
meganle, line 47: can we add docs for the methods like in `campus_eatery.py` for consistency?

src/eatery_db/common_eatery.py
```diff
@@ -1,4 +1,38 @@
 from constants import IMAGES_URL
+from datetime import date, datetime, timedelta
+from database import ExpandedMenuStation, ExpandedMenuItem, ExpandedMenuChoice
+
+today = date.today()
```
meganle: I don't think this should be calculated at the top. It might only get initialized when the server starts but never gets recalculated again. This line should prob be in the functions that use it that way when the function gets called, this variable is recalculated. This might be a possible cause to some of our issues

meganle (changes requested):
Awesome work so far!! Just a few questions

src/db.py
orkosinha: This is because if the data in the CampusEatery table isn't being refreshed, then original `campus_eateries` list has both static and Cornell Dining eateries.  Removing them here should make the list only contain static eateries. I don't think that's a good solution though. I think checking if the eatery is a static eatery makes more sense, so I changed it.

meganle: gotcha

src/eatery_db/campus_eatery.py
```diff
@@ -68,7 +65,7 @@ def parse_campus_hours(data_json, eatery_model):
         eatery_slug = eatery.get("slug")
```
yuna878: should there be a default value (ex. `eatery.get("slug","")` ) in case there's an error?

src/db.py
```diff
@@ -27,40 +35,104 @@ def get_campus_eateries(data_json, refresh=False):
...
+            hours = parse_collegetown_hours(yelp_query, eatery)
+            Session.add_all(hours)
+            Session.commit()
+
     except Exception as e:
         print("Data update failed:", e)
 
 
-start_update()
+start_update(True, True)
```
yuna878: does it need a function call? gunicorn file should take care of the initial call

src/eatery_db/campus_eatery.py
```diff
@@ -132,11 +131,114 @@ def parse_menu_items(items_json, category_model):
...
+                for dates in dates_closed:  # check if dates_closed contains new_date
+                    if "-" in dates:  # indicates this string is a date range of form "mm/dd/yy-mm/dd/yy"
+                        start_date, end_date = string_to_date_range(dates)
+                        if start_date <= new_date <= end_date:
+                            break
+                    else:  # date string is a singular date
+                        closed_date = datetime.strptime(dates, "%m/%d/%y").date()
+                        if new_date == closed_date:
+                            break
+                else:
```
young-k, line 213: i must be blind, or this `else` statement does not line up with any `if` statement.

src/eatery_db/common_eatery.py
```diff
@@ -1,4 +1,36 @@
 from constants import IMAGES_URL
+from datetime import date, datetime, timedelta
+from database import ExpandedMenuStation, ExpandedMenuItem, ExpandedMenuChoice
+
+
+def format_time(start_time, end_time, start_date, hr24=False, overnight=False):
```
young-k: `hr24` is a poor name for this, why not `is_24_hour_time`? or something similar?

src/eatery_db/common_eatery.py
```diff
@@ -25,3 +57,75 @@ def parse_coordinates(eatery):
...
+    json for items
+
+    Args:
+        menu (dict): A valid json segment from the static expanded menu resource.
+        eatery_model (CampusEatery): A valid campus eatery to which to link the menu object.
+    """
+    categories_and_items = []
+
+    for menu in menus["eateries"]:
+        if eatery_model.slug == menu["slug"]:
```
young-k: prefer:

```python
if eatery_model.slug != menu["slug"]:
    break

... # we can keep the indentation level one less
```

src/eatery_db/campus_eatery.py
```diff
@@ -132,11 +131,114 @@ def parse_menu_items(items_json, category_model):
...
+            new_operating_hours = []
+            for i in range(NUM_DAYS_STORED_IN_DB):
+                new_date = today + timedelta(days=i)
+                for dates in dates_closed:  # check if dates_closed contains new_date
+                    if "-" in dates:  # indicates this string is a date range of form "mm/dd/yy-mm/dd/yy"
+                        start_date, end_date = string_to_date_range(dates)
+                        if start_date <= new_date <= end_date:
+                            break
+                    else:  # date string is a singular date
+                        closed_date = datetime.strptime(dates, "%m/%d/%y").date()
```
young-k, line 210: we have this snippet: `datetime.strptime(var_name, "%m/%d/%y")` very frequently, this should be moved into a helper function so that these calls are unified

src/constants.py
```diff
@@ -75,9 +75,18 @@
...
 STATIC_CTOWN_HOURS_URL = STATIC_SOURCES_URL + "externalHours.json"
 STATIC_EATERIES_URL = STATIC_SOURCES_URL + "externalEateries.json"
+STATIC_EATERY_SLUGS = [
+    "Terrace",
+    "Manndible",
+    "Zeus",
+    "Gimme-Coffee",
+    "Fork-and-Gavel",
+    "Louies-Lunch",
+    "Anabels-Grocery",
```
young-k: nit, but alphabetize

young-k (changes requested):
you can dismiss my review if i don't re-review after you fix these

src/eatery_db/campus_eatery.py
meganle: I believe this else clause runs if the break is never hit in the for loop

src/db.py
orkosinha: No, I just have it temporarily to run the function

src/eatery_db/campus_eatery.py
orkosinha: This specific format is only 3 times, is it still worth making a function for it?

src/eatery_db/campus_eatery.py
orkosinha: Added

src/eatery_db/collegetown_eatery.py
orkosinha: Added

src/db.py
meganle: When we start the server do we want to refresh the eateries and recalculate the swipe data? If so, we should prob change the call to `start_update` in `gunicorn.py` to have the `True` args

yuna878: Also, if it's temporary, you should probably delete this line when you merge the changes

orkosinha: Rip, I should've said that I already removed it

meganle (approved):
🚢

young-k (approved):
Changes should be addressed quicker in the future, I don't understand why it took this long to get these in!

PR #157, opened 2019-11-23
Parse from db to graphQL

yuna878:
## Overview

Create parsers that handle querying from db file and reformatting data to match GraphQL schema. Also set up daily cron job (NOTE: set as daily for dev server)

## Changes Made

When a query is made to our graphQL backend, we use the data provided in query (such as specific eatery ids) to query the database containing the eatery information. After fetching data, we reformat the data to match GraphQL object types defined in `src/gql_types` directory. 

## Test Coverage

Made example queries to fetch all possible fields in graphQL and made sure no fields cause error.
Made a same query for local server (db) & deployed server (memory) and kind of confirmed that the return the same result - diff compared the two results and most of the differences were caused because 1) ids of eateries are different between db and memory 2) menu differences between when each server refetched data.

Screen Shot 2019-12-06 at 8 48 11 PM

Screen Shot 2019-12-07 at 4 50 42 PM

### DIFF
--- a/src/gql_parser/campus_eatery.py
+++ b/src/gql_parser/campus_eatery.py
@@ -1,0 +1,328 @@
+from ..constants import SQLITE_MAX_VARIABLE_NUMBER
+from ..db import conn
+from ..database import (
+    CampusEatery,
+    CampusEateryHour,
+    ExpandedMenuChoice,
+    ExpandedMenuItem,
+    ExpandedMenuStation,
+    MenuCategory,
+    MenuItem,
+    SwipeData,
+)
+from ..gql_parser.common_eatery import parse_coordinates, parse_payment_methods, parse_payment_methods_enum
+from ..gql_types import (
+    CampusAreaType,
+    CampusEateryType,
+    EventType,
+    FoodCategoryType,
+    FoodItemType,
+    FoodStationType,
+    DescriptiveFoodItemOptionType,
+    DescriptiveFoodItemType,
+    DescriptiveFoodStationType,
+    OperatingHoursType,
+    SwipeDataType,
+)
+
+
+def get_campus_eateries(eatery_id):
+    """Queries db to fetch information about a specific or all campus eateries.
+
+    Returns a list of CampusEateryType objects.
+    """
+    if eatery_id is not None:
+        query = CampusEatery.query.filter_by(id=eatery_id)
+    else:
+        query = CampusEatery.query
+
+    result = conn.execute(query.statement).fetchall()
+    columns = CampusEatery.__table__.columns.keys()
+
+    populated_result = []
+    for data in result:
+        mapped_eatery = {}
+        for i, column_name in enumerate(columns):
+            mapped_eatery[column_name] = data[i]
+        populated_eatery = parse_campus_eatery(mapped_eatery)
+        populated_result.append(populated_eatery)
+
+    return populated_result
+
+
+def parse_campus_eatery(eatery):
+    """Parses eatery data from db and populates to an object.
+
+    Returns a new CampusEateryType.
+    """
+
+    new_eatery = CampusEateryType(
+        about=eatery.get("about", ""),
+        campus_area=parse_campus_area(eatery),
+        coordinates=parse_coordinates(eatery),
+        eatery_type=eatery.get("eatery_type", ""),
+        expanded_menu=parse_expanded_menu(eatery),
+        id=eatery.get("id"),
+        image_url=eatery.get("image_url"),
+        location=eatery.get("location", ""),
+        name=eatery.get("name", ""),
+        name_short=eatery.get("name_short", ""),
+        operating_hours=parse_operating_hours(eatery),
+        payment_methods=parse_payment_methods(eatery),
+        payment_methods_enums=parse_payment_methods_enum(eatery),
+        phone=eatery.get("phone", "N/A"),
+        slug=eatery.get("slug", ""),
+        swipe_data=parse_swipe_data(eatery),
+    )
+    return new_eatery
+
+
+def parse_campus_area(eatery):
+    """Parses the common name location of an eatery.
+
+    Returns a new CampusAreaType that contains a description of an eatery.
+    """
+    campus_area = eatery.get("campus_area_desc", "")
+
+    return CampusAreaType(description_short=campus_area)
+
+
+def parse_expanded_menu(eatery):
+    """Queries db for expanded menu stations, items, and chocies that are relevant to the specific eatery, then parses
+    the information into appropriate data format.
+
+    Returns a list FoodCategoryType objects.
+    """
+    # query for all menu STATIONs that will be needed to populate expanded menu
+    query_stations = ExpandedMenuStation.query.filter_by(campus_eatery_id=eatery.get("id"))
+    result_stations = conn.execute(query_stations.statement).fetchall()
+    column_stations = ExpandedMenuStation.__table__.columns.keys()
+
+    menu_to_station = {}
+    station_ids = []
+    for station in result_stations:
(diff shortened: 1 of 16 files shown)


src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+    DescriptiveFoodItemType,
+    DescriptiveFoodStationType,
+    OperatingHoursType,
+    SwipeDataType,
+)
+
+
+def get_campus_eateries(eatery_id):
+    if eatery_id is not None:
+        q = CampusEatery.query.filter_by(id=eatery_id)
```
meganle: can we make this variable called `query` instead

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+        for ind in range(len(columns)):
+            mapped_eatery[columns[ind]] = data[ind]
+        populated_eatery = parse_campus_eatery(mapped_eatery)
+        populated_result.append(populated_eatery)
+
+    return populated_result
+
+
+def parse_campus_eatery(eatery):
+    """Parses eatery data from db and populates to an object of type CampusEateryType."""
```
meganle: let's add the return type to the docs here like in the function below

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+    Returns a new CampusAreaType that contains a description of an eatery
+    """
+    campus_area = eatery.get("campus_area_desc", "")
+
+    return CampusAreaType(description_short=campus_area)
+
+
+def parse_expanded_menu(eatery):
+    # query for all menu STATIONs that will be needed to populate expanded menu
+    q_stations = ExpandedMenuStation.query.filter_by(campus_eatery_id=eatery.get("id"))
```
meganle: maybe `station_query`? try to avoid single letters in variables unless it's for like an index in a loop

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+        menu_category = mapped_station["menu_category"]
+        if menu_category in menu_to_station:
+            menu_to_station[menu_category].append(mapped_station)
+        else:
+            menu_to_station[menu_category] = [mapped_station]
+        station_ids.append(mapped_station["id"])
+
+    # query for all menu ITEMs that will be needed to populate expanded menu
+    result_items = []
+    while len(station_ids) > 999:
```
meganle: what is this while condition supposed to mean?

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+    # query for all menu STATIONs that will be needed to populate expanded menu
+    q_stations = ExpandedMenuStation.query.filter_by(campus_eatery_id=eatery.get("id"))
+    result_stations = conn.execute(q_stations.statement).fetchall()
+    column_stations = ExpandedMenuStation.__table__.columns.keys()
+
+    menu_to_station = {}
+    station_ids = []
+    for station in result_stations:
+        mapped_station = {}
+        for ind in range(len(column_stations)):
```
meganle: You can do something like
`for i, station in enumerate(column_stations):` and then on the next line `mapped_station[station] = station[i]`

meganle: `enumerate` is really helpful if you need both the index and the item, and it makes code more readable

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+    else:
+        q = CampusEatery.query
+
+    result = conn.execute(q.statement).fetchall()
+    columns = CampusEatery.__table__.columns.keys()
+
+    populated_result = []
+    for data in result:
+        mapped_eatery = {}
+        for ind in range(len(columns)):
```
meganle: like enumerate would be good here also! check comment below

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+        partial_result = conn.execute(partial_query.statement).fetchall()
+        result_items += partial_result
+        station_ids = station_ids[999:]
+    column_items = ExpandedMenuItem.__table__.columns.keys()
+
+    station_to_item = {}
+    item_ids = []
+    for item in result_items:
+        mapped_item = {}
+        for ind in range(len(column_items)):
```
meganle: same with enumerate

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+        exists = station_to_item[station_category]
+        if exists is not None:
+            station_to_item[station_category].append(mapped_item)
+        else:
+            station_to_item[station_category] = [mapped_item]
+        item_ids.append(mapped_item["id"])
+
+    # query for all menu ITEMs that will be needed to populate expanded menu
+    result_choices = []
+    while len(item_ids) > 999:
```
meganle: 999 is used a lot here so we should prob put it in a constant either at the top of this file or `constants.py`

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+                start_time=event["start_time"],
+            )
+            events_objs_arr.append(events_obj)
+
+        result_obj = OperatingHoursType(date=date, events=events_objs_arr)
+        populated_result.append(result_obj)
+
+
+def parse_swipe_data(eatery):
+    q = SwipeData.query.filter_by(eatery_id=eatery.get("id"))
```
meganle: query

src/gql_parser/campus_eatery.py
```diff
@@ -0,0 +1,295 @@
...
+    result = conn.execute(q.statement).fetchall()
+    columns = SwipeData.__table__.columns.keys()
+
+    populated_result = []
+    for data in result:
+        mapped_swipe_data = {}
+        for ind in range(len(columns)):
```
meganle: i'll stop pointing out the enumerate stuff LOL i think u get the gist

src/gql_parser/collegetown_eatery.py
```diff
@@ -0,0 +1,107 @@
...
+    CollegetownHoursType,
+    PaymentMethodsType,
+    PaymentMethodsEnum,
+    RatingEnum,
+)
+
+
+def get_collegetown_eateries(eatery_id):
+    if eatery_id is not None:
+        q = CollegetownEatery.query.filter_by(id=eatery_id)
```
meganle: query

src/gql_parser/collegetown_eatery.py
```diff
@@ -0,0 +1,107 @@
...
+    return RatingEnum.get(index)
+
+
+def parse_categories(eatery):
+    categories = eatery.get("categories", "").split(",")
+    return list(filter(lambda category: category != "", categories))
+
+
+def parse_collegetown_hours(eatery):
+    q = CollegetownEateryHour.query.filter_by(eatery_id=eatery.get("id"))
```
meganle: i will also stop pointing out `query` LOL

meganle (changes requested):
whew that's a thick PR but amazing job so far!

PS You're def doing enough work for appdev

src/gql_parser/campus_eatery.py
yuna878: SQLite doesn't accept more than 999 variables in a single query. Doing the `filter..in` thorws errors because `station_ids` would exceed the length so I run a while loop to break up the query into multiple smaller queries and aggregate the result into a single array.

meganle (approved):
Let's goooo 🐟

orkosinha (approved):
lgtm!

PR #158, opened 2019-12-01
Add fix for dates to release

orkosinha:
## Overview

Previously the `today` variable stored the date, but it was only initialized when the code was first executed. This should fix that so the date is accurate to when the code is executed rather than the date of the first execution. 

## Changes Made

Initialize the `today` variable in every function call rather than as a global variable.
I also didn't do this through master because the master currently has a different `gunicorn.py` file which calls the db parsing methods rather than the current ones, so this fix is only for release.

This change is made in `static_eatery.py` `collegetown_eatery.py` and `common_eatery.py`

I think I messed up my branches and this PR also includes #155


johnathananderson (approved):

yuna878 (approved):
looks good!

seancorc (approved):
Looks good to me!

PR #167, opened 2019-12-08
Add installation for cron in Dockerfile

meganle:
## Overview
Since we use crontab to update the DB daily, the Docker image needs this dependency

## Changes Made

Add installation script to Dockerfile


kevinchan159 (approved):
I love crone-jobs 🤗

yuna878 (approved):
crone master

PR #168, opened 2019-12-08
Edit cronjob to update multiple times a day

meganle:
## Overview
Edit cronjob to run DB update at 4 am, 10 am, and 4 pm eastern time everyday 

## Changes Made

Edit cronjob to run at 4 am, 10 am, and 4 pm eastern time everyday to make sure menus are up to date for lunch and dinner since Cornell dining doesn't have a consistent time where they update their menus


kevinchan159 (approved):
crone-job master

PR #169, opened 2019-12-08
Fix parsing and DB storage of operating hours

meganle:
## Overview
Noticed the operating hours on prod and the new DB setup weren't matching up. On the new DB setup if an eatery (collegetown and campus) was closed, there was no entry for that day even though prod had an entry. Also, I noticed that the Lite Dinners with no menus appeared again in the DB setup when in the prod it was just merged with the previous event. It's cause the call to `merge_hours` was never called.

## Changes Made
Made sure we always created entries for eateries even if they were closed for consistency. This required changing fields in `CollegetownEateryHour` to be nullable which actually is now consistent with `CampusEateryHour`. And to make sure the Lite Dinners and Lite Lunches were properly merged, I added the `merge_hours` call and copied the function from `data.py` since that file will be deprecated soon.

## Test Coverage
I diffed the output from production with my local environment and it all matched up. The only issue now is with Mac's hours which will be addressed next.

### DIFF
--- a/src/database/CollegetownEateryHour.py
+++ b/src/database/CollegetownEateryHour.py
@@ -10,6 +10,6 @@ class CollegetownEateryHour(Base):
 
     id = Column(Integer, nullable=False, primary_key=True)
     date = Column(String, nullable=False)
-    event_description = Column(String, nullable=False)
-    end_time = Column(String, nullable=False)
-    start_time = Column(String, nullable=False)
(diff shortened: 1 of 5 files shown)


orkosinha (approved):
lgtm, but are terrace hours also not showing up or just Macs?

yuna878 (commented):
💯

yuna878 (approved):

PR #170, opened 2019-12-08
Add macs to static slugs

meganle:
## Overview

Earlier there was an issue where the hours for Mac's weren't showing up. I found that Mac's was a static eatery and that its slug was not included in a necessary constant. Because of this, the hours for Mac's was not parsed correctly

## Changes Made
Added Mac's slug to the list of static slugs in `constants.py`

## Test Coverage
I diffed the Mac's info on the prod server with my local server and found them to match exactly


kevinchan159 (approved):
macdonalds

orkosinha (approved):
lgtm

PR #171, opened 2019-12-09
Fix missing swipe data for Terrace

meganle:
## Overview
Swipe data was empty for Terrace but on prod it was present. I found out that this stems from the fact that Terrace is considered a static eatery. Previously, when adding swipe data to the DB, it would only add swipe data for non-static campus eateries. This fixes that to include static campus eateries.

## Changes Made
I moved the swipe data calculation to after we parse both campus and static eateries. Then, I also passed in static eateries to the `export_data` function so Terrace will be recognized when adding swipe data to the DB.

## Test Coverage
I tested locally and made sure that the swipe data information provided matched the prod server

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -43,15 +43,6 @@ def start_update(refresh_campus=False, recalculate_swipe=False):
         campus_json = requests.get(CORNELL_DINING_URL).json()
         campus_eateries = get_campus_eateries(campus_json, refresh=refresh_campus)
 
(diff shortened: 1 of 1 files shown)


kevinchan159 (approved):
🤯

orkosinha (approved):
lgtm

yuna878 (approved):
😃

PR #176, opened 2019-12-11
Fix static eatery category

yuna878:
## Overview

Fix static eatery's menu categories from `General` to empty string to make it consistent with old logic


meganle (approved):
dok dok

PR #177, opened 2019-12-11
Fix date to use Eastern timezone

meganle:
## Overview

Previously, we used `date.today()` to get the current date. However, this uses the local timezone and our servers use UTC time which is 5 hours ahead. This sometimes causes us to mistakenly think that the current date is actually one day ahead.

## Changes Made

Added function to get the current date in Eastern timezone. I added this to `constants.py` since there is no other file for helper functions.

### DIFF
--- a/src/constants.py
+++ b/src/constants.py
@@ -1,3 +1,6 @@
+from datetime import datetime
+import pytz
+
 ACCOUNT_NAMES = {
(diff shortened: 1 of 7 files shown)


yuna878 (approved):
🔥

kevinchan159 (approved):
looks 👍 2me

PR #181, opened 2020-01-27
Support new BRB accounts

orkosinha:
A really small fix to add BRB support for Spring semester.

## Overview

Since the `"accountName"` field changed the BRB label from `"BRB Big Red Bucks"` to `"BRB Big Red Bucks Spring"` and we used an explicit equals, new BRB values were not being recognized. This fix should ideally also work in the Fall depending on how Get chooses to rename the BRB accounts.


whoiswillma (approved):

young-k (approved):

alanna-zhou (approved):
wau

PR #183, opened 2020-01-28
Change script field and add to SSH host for Travis

meganle:
## Overview
Travis updated some of their specs in regard to the yml file and currently everything doesn't pass.

## Changes Made
The script field is required to have something in it so I just added an `echo` call to make it pass. In addition, Travis is not allowing you to put the host name and suggests using the public key of the server instead (https://docs.travis-ci.com/user/ssh-known-hosts/). Lastly, the skip_cleanup field changed to just cleanup.

## Test Coverage
Travis currently passes on my branch

Changing this for the other backends

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,13 @@
 language: python
 python:
 - '3.6'
-sudo: required
 services:
 - docker
-script:
-- 
-addons:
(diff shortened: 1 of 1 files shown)


yuna878 (approved):
I believe in you :)

meganle:
@aliciaxw Remind me to show you how to do this for the other repos!

PR #184, opened 2020-01-28
Fix cronjob to run in container

meganle:
## Overview
There's some extra steps required to run a cronjob within a Docker container.

## Changes Made
Edited Dockerfile to actually run cronjob when the container starts and also added logging so we can verify

## Test Coverage
Tested by running locally


yuna878 (approved):
Finally :D

orkosinha (approved):
looks fire

PR #188, opened 2020-02-01
Clean up unused code

yuna878:
## Overview

Deleted a file & function that are no longer in use after DB transition.

## Test Coverage

Ran it locally to make sure the deletion doesn't cause any errors.


meganle (approved):
🐟

leewang-codes (approved):
🐟

PR #192, opened 2020-02-14
Quick fix for campus updates

yuna878:
## Overview

Fix so that campus updates doesn't erase collegetown eateries

### DIFF
--- a/src/db.py
+++ b/src/db.py
@@ -3,7 +3,21 @@ import requests
 
 from .collegetown import collegetown_search
 from .constants import CORNELL_DINING_URL, STATIC_EATERIES_URL, STATIC_EATERY_SLUGS, STATIC_EXPANDED_ITEMS_URL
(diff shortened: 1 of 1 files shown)


meganle (approved):

src/db.py
```diff
@@ -25,8 +39,32 @@
 
 def get_campus_eateries(data_json, refresh=False):
     if refresh:
-        Base.metadata.drop_all(bind=Engine)
-        Base.metadata.create_all(bind=Engine)
+        Base.metadata.drop_all(
```
whoiswillma, line 42: Could I get a summary on how this fixes the item duplication issue we just saw?

Also, is this fix a temporary one?

yuna878: Before the change, db tables that were unrelated to the corresponding update operations were being modified so this fix makes sure that only the tables directly related to the operation is modified.

leewang-codes (approved):

PR #193, opened 2020-02-22
Remove cron job for ctown

meganle:
## Overview
Currently, ctown tries to update every week but cron doesn't have access to the required env vars it needs due to some weird stuff between cron and Docker. It's not necessary to update ctown eateries once a week since they rarely change and will update upon startup.

## Changes Made
- Remove ctown update job


yuna878 (approved):
🔥

PR #200, opened 2020-08-23
Update externalEateries.json

whoiswillma:
Update Louie's lunch since it's open again ???


leewang-codes (approved):
it's open again!? bless

alanna-zhou (approved):
😮

PR #212, opened 2020-10-20
Add eatery types

alanna-zhou:
## Overview
Add eatery types because they return unknown right now from the static json file

## Test Coverage
Will be deploying server at night to see how it goes


chalo2000 (approved):
LGTM

static_sources/externalEateries.json
```diff
@@ -774,4 +813,4 @@
             ]
         }
     ]
-}
+}
```
whoiswillma: Could you try adding a newline here to get the check to pass?

PR #213, opened 2020-10-28
Integrate OpenTable/Get

raahimenon:
## Overview
Added fields to the database to support querying about reservation links (currently Cornell uses OpenTable) and whether the eatery uses Get.

## Changes Made
If you add the `reserveLink` query to `campusEateries`, you will get the link allowing you to reserve space at that eatery, if such a link exists. Cornell uses OpenTable now and for the foreseeable future, but I wanted to make the field name more general just in case. Adding the `isGet` query will return a boolean value for whether or not that eatery is integrated with Get.

## Test Coverage
This was all done through manual testing in the GraphQL query engine.

I'm not entirely sure if I managed to get every link/get dining hall; I found all the ones I could but if I made a mistake the JSON will need to be updated.

## Misc
I managed to mess up the tense of my commit messages again; I'll try to switch to present tense in the future.


src/constants.py
```diff
@@ -80,7 +80,7 @@
     "summer": "5/17/20-8/27/20",
 }
 SQLITE_MAX_VARIABLE_NUMBER = 999
-STATIC_SOURCES_URL = GIT_CONTENT_URL + "/eatery-backend/master/static_sources/"
+STATIC_SOURCES_URL = GIT_CONTENT_URL + "/eatery-backend/raahi/openTable-get/static_sources/"
```
alanna-zhou, line 83: why does this include your name

src/constants.py
```diff
@@ -95,7 +95,7 @@
 ]
 STATIC_EXPANDED_ITEMS_URL = STATIC_SOURCES_URL + "expandedItems.json"
 STATIC_MENUS_URL = STATIC_SOURCES_URL + "hardcodedMenus.json"
-STATIC_EXCEPTIONS_URL = STATIC_SOURCES_URL + "exceptions.json"
+STATIC_ATTRIBUTES_URL = STATIC_SOURCES_URL + "eateryAttributes.json"
```
alanna-zhou, line 98: why is it not just `attributes.json`

src/database/CampusEatery.py
```diff
@@ -23,3 +23,5 @@ class CampusEatery(Base):
     phone = Column(String, nullable=False)
     slug = Column(String, nullable=False)
     exceptions = Column(String, nullable=False)
+    reserve_link = Column(String, nullable=True)
```
alanna-zhou, line 26: `reserve_url` might be better?

src/gql_types/eatery_type.py
```diff
@@ -1,4 +1,4 @@
-from graphene import Enum, Field, Float, Int, List, ObjectType, String
+from graphene import Enum, Field, Float, Int, List, ObjectType, String, Boolean
```
alanna-zhou, line 1: alphabetize this import

static_sources/eateryAttributes.json
```diff
@@ -0,0 +1,67 @@
+{
+  "104-West": {
+    "reserveLink": "https://www.opentable.com/restaurant/profile/1080703"
```
alanna-zhou, line 3: by convention we usually would do something like `reserveUrl`

static_sources/eateryAttributes.json
```diff
@@ -0,0 +1,67 @@
...
+  "Bear-Necessities": {
+    "exceptions": [
+      "Mobile Order Only"
+    ],
+    "isGet": true
+  },
+  "Becker-House-Dining": {
+    "reserveLink": "https://www.opentable.com/restaurant/profile/1080700"
+  },
+  "Bug-Stop-Bagels": {
```
alanna-zhou, line 14: i didn't know there was a Bug Stop Bagels on campus :O

static_sources/eateryAttributes.json
```diff
@@ -0,0 +1,67 @@
...
+  "Carols-Cafe": {
+    "isGet": true
+  },
+  "Cook-House-Dining": {
+    "reserveLink": "https://www.opentable.com/restaurant/profile/1080697"
+  },
+  "Cornell-Dairy-Bar": {
+    "isGet": true
+  },
+  "Goldies-Cafe": {
```
alanna-zhou, line 29: do these `Cafe`s not need accent marks on the e?

static_sources/eateryAttributes.json
```diff
@@ -0,0 +1,67 @@
+{
+  "104-West": {
```
alanna-zhou, line 2: is there a reason why we can't do `"104 West"` and `reserve_link`? like is iOS expecting this kind of text format cuz it's kinda weird

alanna-zhou (changes requested):
looks great!! let's address comments first

yuna878 (approved):
Agree with all of Alanna's comments! Good work :)

src/constants.py
raahimenon: oh yikes I forgot to change the url back to master thanks

src/constants.py
raahimenon: also a good question I think I just saw that the other files had the word eatery in them and wanted to follow suit, but it definitely doesn't need it; I'll change it

src/database/CampusEatery.py
raahimenon: ah yeah you're right

src/gql_types/eatery_type.py
raahimenon: will do

static_sources/eateryAttributes.json
raahimenon: see I didn't either, but this is BSB's official slug from Cornell - I checked the Cornell Dining JSON just to be sure

static_sources/eateryAttributes.json
raahimenon: nope, slugs remove any weird characters

static_sources/eateryAttributes.json
raahimenon: the indices in the attributes JSON are from the eateries slugs, which we already have stored as a convenient unique identifier - I didn't want to use the name of the eatery because I didn't really want to deal with accent marks. I should probably add a file somewhere that has a list of slugs for each eatery at some point, but for now it's easy enough to find by querying for the slug field in GraphQL (that's how I generally found them)

raahimenon: as for the second, I was looking up JSON code practices and found somewhere that camelCase was kind of preferred? It was a random stackoverflow post though so I can just change it to snake_case

static_sources/eateryAttributes.json
raahimenon: yeah that's my bad I just used the first non OpenTable variable name I thought of

PR #214, opened 2020-10-31
OpenTable / Get PR Refactor

raahimenon:
## Overview
Refactored the old PR to use the word 'url' instead of 'link'. Made variables snake_case instead of camelCase (GraphQL queries are still camelCase). Fixed problem where external eateries wouldn't update. Alphabetized imports.  

## Changes Made
This was mostly variable name changes and rearranging code. One important thing to note is the GraphQL query has changed from "reserveLink" to "reserveUrl". "isGet" is still the correct query for whether or not an eatery is accessible through Get. I'm a little on the fence about whether the variable names should actually be snake_case in the JSON file, since GraphQL forces queries to be camelCase, and if someone from a different subteam were to edit the file they would have to know to convert between conventions, but for now this is fine.

## Test Coverage
Manual testing to make sure the DB updated properly.

## Query Fields:
reserveLink -> reserveUrl
isGet -> isGet


alanna-zhou (approved):
looks great but you should update your PR title to something like `Refactor OpenTable and Get` -- let's make this the last time that either Yuna and I request this! xD

raahimenon:
> looks great but you should update your PR title to something like `Refactor OpenTable and Get` -- let's make this the last time that either Yuna and I request this! xD

Wait I chose this title because I thought it counted as present tense but good to know it should start with a verb - I think I got this from now on!

PR #218, opened 2020-11-23
Update Terrace Exception

raahimenon:
## Overview
Terrace is closed at the moment, so an exception was added to note that.


raahimenon:
It was a temporary thing; they're back open now

PR #219, opened 2020-11-29
Update externalEateries.json

whoiswillma:
From Jana:

> One change that we have is that after Thanksgiving, we will be Monday – Friday.


leewang-codes (approved):
lgtm

PR #220, opened 2020-11-29
Remove bigger menu category for expanded items

alanna-zhou:
## Overview
Remove menu category -- apparently it's not needed

Example GraphQL query:
```
{
  eateries {
    name,
    expandedMenu {
      category,
      stations {
        items {
          item
          choices {
            options
          }
        }
      }
    }
  }
}
```

## Test Coverage
Works locally

You can also verify yourself by modifying `db.py` with
```
with open(f"{os.getcwd()}/static_sources/expandedItems.json") as f:
                data = json.load(f)
menus = data  # requests.get(STATIC_EXPANDED_ITEMS_URL).json()
```

## Next Steps 
Make sure it deploys on server nicely

## Screenshots

<details>

  <summary>Before</summary>

Screen Shot 2020-11-29 at 1 28 47 PM  

</details>

<details>

  <summary>After</summary>

Screen Shot 2020-11-29 at 1 29 02 PM

</details>


leewang-codes (approved):
rip menu category

yuna878 (approved):
🔥

whoiswillma:
Just to be clear, this is only to do with expanded menu items right? On the frontend for regular menu items we still very much depend on the category for dining halls a la Becker and Appel. 

@spd75 can weigh in whether he's started connecting iOS up with the backend, but it should be non-breaking since the live version doesn't depend on anything to do with expanded menu items.

alanna-zhou:
> Just to be clear, this is only to do with expanded menu items right? On the frontend for regular menu items we still very much depend on the category for dining halls a la Becker and Appel.
> 
> @spd75 can weigh in whether he's started connecting iOS up with the backend, but it should be non-breaking since the live version doesn't depend on anything to do with expanded menu items.

yes, only expanded menu items

hockeymonday (approved):

chalo2000 (approved):
LGTM, was a little confused about why we had a property of station in the expanded_menu and then another one outside of the menu but still in a tuple along with it, but looks like it's in a parse function so I'm assuming it's strictly required in that form somewhere else

PR #221, opened 2020-12-12
Update externalEateries.json

whoiswillma:
From Shelby: 

> The last day that we will be open is Wednesday, December 16.

After Wednesday I'll update it so that Terrace is permanently closed


alanna-zhou (approved):
nice

PR #224, opened 2020-12-16
Update externalEateries.json

whoiswillma:
From Jana:

> Terrace has closed early today due to the weather.
> This is also our last day of 2020.
> We expect to open again on Saturday, January 23, 11:00am – 6:30pm.
> If there are any changes closer to opening, I will send an updated email.


connorreinhold (approved):
Oh. This makes sense 👍

PR #234, opened 2021-02-21
Fix 104 West! OpenTable URL

raahimenon:
## Overview
104 West! url somehow directed to a Trillium passover menu, so that's been fixed.


grebneerg (approved):
Looks good!

PR #235, opened 2021-02-21
Fix Expanded Menus Associated with all Eateries

raahimenon:
## Overview
Expanded Menus were all assigned to every eatery; now they act as intended with their respective eateries.


raahimenon:
accidental closure

grebneerg (approved):
Looks good

PR #243, opened 2021-05-01
Close Gimme Coffee

whoiswillma:
https://gimmecoffee.com/locations/

Doesn't explicitly say that Gates is closed, but I've (a) heard through the grapevine that it is and (b) you can't order online from the delivery page.


alanna-zhou (approved):
nice

PR #244, opened 2021-05-02
update expandedItems.json up to Gimme Coffee

raahimenon:
Added menus to `expandedItems.json` as specified in the dining doc up to "Gimme Coffee."


alanna-zhou (approved):
`Update expandedItems.json up to Gimme Coffee` not `update` haha

PR #253, opened 2021-08-26
Update externalEateries.json

whoiswillma:
- Update attributes.json
  - Close Manndibles (RIP)
  - Set temple of zeus until 5PM (saw this in person)


raahimenon (approved):
Ty will

PR #257, opened 2021-09-17
Update externalEateries.json

whoiswillma:
The website says until 5:30PM, https://gimmecoffee.com/locations/

But I think in person it says until 3PM. I will err on the side of caution in this case.

### DIFF
--- a/static_sources/externalEateries.json
+++ b/static_sources/externalEateries.json
@@ -498,7 +498,19 @@
                     "descrshort": "cafe"
                 }
             ],
-            "operatingHours": [],
+            "operatingHours": [
+                {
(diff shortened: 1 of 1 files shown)


alanna-zhou (approved):
is there any way that @snajima could have caught this on his own? trying to transition as much as possible away from will ma

alanna-zhou:
> is there any way that @snajima could have caught this on his own? trying to transition as much as possible away from will ma

in the future could you ping shungo to make these changes? at least you can guide him on what to do + i'd prefer you not making any more pull requests as much as i appreciate you (and as much as you think these are trivial)

whoiswillma:
Sure

PR #262, opened 2021-11-27
Update externalEateries.json

whoiswillma:
> Both outlets reopen with regular hours on Monday, November 29.


leewang-codes (approved):
lgtm

